PR #41, opened 2023-04-04
Move to independent pre-commit checks

### LINKED ISSUE #40: Improve performance times by exploring decoupled pre-commit implementation
Testing performance times may likely be improved by decoupling pre-commit from other tests. It could be that specific linting is more effectively performed from within the Poetry environment as well.

d33bs:
# Description

This change moves pre-commit checks away from Dagger-based tests in an effort to improve testing performance times and take further advantage of GitHub Actions workflows. Pre-commit manages environments for hooks independent of Poetry by default (despite the ability to install pre-commit into Poetry environments), so these changes also are an attempt to use pre-commit as designed.

Other changes:
- Dependencies were updated where possible to prepare for other changes needed to in this project.
- In updating dependencies, a minor test update was needed.

Please let me know if you have any questions or concerns, and thank you for your feedback!

Closes #40 

- [x] This change requires a documentation update.

# Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -66,14 +66,11 @@ repos:
         # pylint due to it's dynamic analysis capabilities
         # and the way pre-commit references virtual environments.
         additional_dependencies:
-          - "cloudpathlib[all]>=0.12.0"
+          - "cloudpathlib[all]>=0.13.0"
           - "pyarrow>=11.0.0"
           - "prefect>=2.7.0"
-          - "prefect-dask>=0.2.2"
+          - "prefect-dask>=0.2.3"
           - "pytest>=7.2.0"
-          - "moto[server,s3]>=4.0.11"
-          - "duckdb>=0.6.0"
-          - "git+https://github.com/cytomining/pycytominer.git@afac3ea"
-        args:
-          # adding this arg as advised by pylint:I1101
-          - --extension-pkg-allow-list=duckdb
+          - "moto[server,s3]>=4.0.0"
+          - "duckdb>=0.7.0"
+          - "git+https://github.com/cytomining/pycytominer.git@c90438fd7c11ad8b1689c21db16dab1a5280de6c"
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,5 +1,7 @@
 # See https://pre-commit.com for more information
 # See https://pre-commit.com/hooks.html for more hooks
+default_language_version:
+  python: python3.10
 repos:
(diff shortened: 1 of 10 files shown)


.pre-commit-config.yaml
```diff
@@ -66,14 +66,11 @@ repos:
...
           - "pytest>=7.2.0"
-          - "moto[server,s3]>=4.0.11"
-          - "duckdb>=0.6.0"
-          - "git+https://github.com/cytomining/pycytominer.git@afac3ea"
-        args:
-          # adding this arg as advised by pylint:I1101
-          - --extension-pkg-allow-list=duckdb
+          - "moto[server,s3]>=4.0.0"
+          - "duckdb>=0.7.0"
+          - "git+https://github.com/cytomining/pycytominer.git@c90438fd7c11ad8b1689c21db16dab1a5280de6c"
```
gwaybio, line 76: I think an official 0.3 pycytominer release is in our near future! 😁

gwaybio (approved):

.pre-commit-config.yaml
falquaddoomi, line 76: *(Copied from another PR, but it should be here AFAICT)*

So, my system Python on my M1 mac is 3.11.2, and pre-commit uses the system Python version when it attempts to install dependencies. Apparently scipy version 1.6.1 is the most recent version of scipy that doesn't exclude Python 3.11( which is likely an oversight on the package maintainer's part). Unfortunately, this release is built from source, and isn't compatible with the version of openblas/lapack that's installed by default via brew on my machine. I didn't really trust my ability to go through the various versions and installation options for those packages, especially considering that installing pycytominer at that commit in a Python 3.10 virtualenv installs scipy 1.8.1 (since it just excludes anything <=3.11) and works without a hitch.

A quick solution would be to specify the Python version you support as the default, i.e. by adding this as a top-level attribute:
```
default_language_version:
  python: python3.10
```
After adding that to `.pre-commit-config.yaml` I was able to run `pre-commit run --all-files` successfully. Change it to whichever version of Python you're targeting, of course, but in my experience 3.11 still seems to have issues.

falquaddoomi (commented):
Looks good aside from my difficulties in running pre-commit on Python 3.11.2. I made a comment about the problem as well as a potential resolution.

.pre-commit-config.yaml
d33bs: Thanks for this @falquaddoomi ! I'll add this in as a configuration for pre-commit.

falquaddoomi (approved):

Issue #27, opened 2023-01-23
Explore the use of schema for input or output specification


d33bs:
This issue focuses on exploring and possibly implementing the use of schema as a way to describe input or output data. Specifically, this would be a way to describe column names, data types, and groupings of files using an existing standard of some kind.

PR #42, opened 2023-04-06
Enable column name compatibility with Pycytominer results

### LINKED ISSUE #39: Address column naming differences from Pycytominer-processed CellProfiler SQLite sources
Currently there is a gap between Pycytominer processed SQLite files (using `SingleCells.merge_single_cells()`) and CytoTable (using `convert` functionality). This gap occurs where the string "Metadata_" is added to specific columns. A configuration preset change in addition to other aspects may need to be addressed.

Specifically, the following columns are named differently:

<table>
<tr>
<td>
Pycytominer columns which are named differently from CytoTable results on same dataset</td>
<td> CytoTable columns which are named differently from Pycytominer results on same dataset</td>
</tr>
<tr>
<td>

```python
['Metadata_Cytoplasm_Parent_Cells',
 'Metadata_Cytoplasm_Parent_Nuclei',
 'Metadata_Cells_Number_Object_Number',
 'Metadata_Nuclei_Number_Object_Number']
```
</td>
<td>

```python
['Cytoplasm_Parent_Cells',
 'Cytoplasm_Parent_Nuclei',
 'Cells_Number_Object_Number',
 'Nuclei_Number_Object_Number']
```

</td>
</tr>
</table>

d33bs:
# Description

This change addresses minor column name differences between Pyctyominer `SingleCells.merge_single_cells` and CytoTable `convert` results (outlined on #39). A test-driven approach was taken to ensure alignment is accounted for with this and future changes.

Thank you in advance for your thoughts and feedback on this PR!

Note: this builds on changes from #41.

Closes #39 

# Checklist

### DIFF
--- a/cytotable/convert.py
+++ b/cytotable/convert.py
@@ -94,7 +94,7 @@ def _prepend_column_name(
             Name of data source source group (for common compartments, etc).
         identifying_columns: Union[List[str], Tuple[str, ...]]:
             Column names which are used as ID's and as a result need to be
-            ignored with regards to renaming.
+            treated differently when renaming.
         metadata: Union[List[str], Tuple[str, ...]]:
             List of source data names which are used as metadata
         targets: List[str]:
@@ -126,7 +126,7 @@ def _prepend_column_name(
 
     for column_name in source["table"].column_names:
         # if-condition for prepending source_group_name_stem to column name
-        # where colname is not an identifying column
+        # where colname is not in identifying_columns parameter values
         # and where the column is not already prepended with source_group_name_stem
         # for example:
         #   source_group_name_stem: 'Cells'
@@ -137,7 +137,7 @@ def _prepend_column_name(
         ):
             updated_column_names.append(f"{source_group_name_stem}_{column_name}")
         # if-condition for prepending 'Metadata_' to column name
-        # where colname is an identifying column
+        # where colname is in identifying_columns parameter values
         # and where the column is already prepended with source_group_name_stem
         # for example:
         #   source_group_name_stem: 'Cells'
@@ -148,7 +148,7 @@ def _prepend_column_name(
         ):
             updated_column_names.append(f"Metadata_{column_name}")
         # if-condition for prepending 'Metadata' and source_group_name_stem to column name
-        # where colname is an identifying column
+        # where colname is in identifying_columns parameter values
         # and where the colname does not already start with 'Metadata_'
         # and colname not in metadata list
         # and colname does not include 'ObjectNumber'
@@ -167,7 +167,7 @@ def _prepend_column_name(
             )
         # if-condition for prepending 'Metadata' to column name
         # where colname doesn't already start with 'Metadata_'
-        # and colname is in identifying columns list
+        # and colname is in identifying_columns parameter values
         # for example:
         #   column_name: 'ObjectNumber'
         #   updated_column_name: 'Metadata_ObjectNumber'
--- a/cytotable/convert.py
+++ b/cytotable/convert.py
@@ -136,7 +136,17 @@ def _prepend_column_name(
             source_group_name_stem.capitalize()
         ):
             updated_column_names.append(f"{source_group_name_stem}_{column_name}")
-
+        # if-condition for prepending 'Metadata_' to column name
+        # where colname is an identifying column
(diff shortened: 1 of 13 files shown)


cytotable/convert.py
```diff
@@ -133,10 +133,20 @@ def _prepend_column_name(
...
         ):
             updated_column_names.append(f"{source_group_name_stem}_{column_name}")
-
+        # if-condition for prepending 'Metadata_' to column name
+        # where colname is an identifying column
+        # and where the column is already prepended with source_group_name_stem
+        # for example:
+        #   source_group_name_stem: 'Cells'
+        #   column_name: 'Cells_Number_Object_Number'
+        #   updated_column_name: 'Metadata_Cells_Number_Object_Number'
```
gwaybio, line 145: Will this prepend `Metadata` to columns that shouldn't have this prefix? E.g. will `Cells_AreaShape` become `Metadata_Cells_AreaShape`? We want to avoid this.

gwaybio: the documentation is already stellar, but if my concern is not applicable, please add a brief note describing this.

cytotable/presets.py
```diff
@@ -92,6 +92,54 @@
                 AND per_nuclei.Nuclei_Number_Object_Number = per_cytoplasm.Cytoplasm_Parent_Nuclei
             """,
     },
+    "cellprofiler_sqlite_pycytominer": {
```
gwaybio, line 95: love this!

Could it also be worth specifying `"CELLPROFILER_VERSION"` as a field? I think some of the identifying columns may be different. CC'ing @jenna-tomkinson to hear her thoughts on any differences she may forsee according to CellProfiler version

tests/test_convert.py
```diff
@@ -734,3 +736,78 @@ def test_convert_dask_cellprofiler_csv(
 
     assert test_result.shape == control_result.shape
     assert test_result.equals(control_result)
+
+
+def test_convert_cellprofiler_sqlite_pycytominer_merge(
```
gwaybio, line 741: this is so elegant 🤩

gwaybio (approved):
very minor comments that I don't think should slow merge

cytotable/presets.py
jenna-tomkinson: I would hope that CellProfiler does not change the naming conventions of the columns anytime in the future. It would be very unnecessary unless there was a big reason why it would need to be changed.

It might be good to add the CellProfiler version in general just in case.

cytotable/convert.py
d33bs: Thanks! The key with this if branch is that the column name must be an "identifying column" which is provided upfront as a parameter to `convert` via `identifying_columns`. In the example you provided, `Cells_AreaShape` would only ever be renamed that way if it was specified as an identifying column by a developer. Configuration preset examples can be found under the reused key name `CONFIG_IDENTIFYING_COLUMNS` under [`presets.config`](https://github.com/cytomining/CytoTable/blob/main/cytotable/presets.py). You're right in mentioning I should be more upfront about this in the documentation, I'll see about highlighting this better.

cytotable/presets.py
d33bs: Thank you both! I like this idea and will work towards incorporating it in.

tests/test_convert.py
d33bs: Cheers, thanks!

cytotable/presets.py
d33bs: I've added this data into the configuration presets. In the case of the configuration preset used for this PR's addition, I felt that we needed to reference both CellProfiler and Pycytominer's versions, so the new key is itself a dictionary of multiple elements to help keep this flexible. Note: I also intend to add more comprehensive schema details for input and output in #27, which I feel will complement this work.

falquaddoomi (commented):
This all looks good as far as I can tell, but I had an issue with pre-commit attempting to install scipy 1.6.1 (when it should have installed 1.8.1) and failing as a result. I made a suggestion for how to potentially fix that for others.

falquaddoomi (approved):

d33bs:
Thank you @gwaybio , @jenna-tomkinson and @falquaddoomi for your reviews here! I've made changes to address the comments (and have updated this branch from main). Merging shortly.

Issue #43, opened 2023-04-10
Invalid Input Error


gwaybio:
Hi @d33bs - I am running into the following error when testing `cytotable.convert()`

```python
duckdb.InvalidInputException: Invalid Input Error: Attempting to execute an unsuccessful or closed pending query result
Error: Invalid Error: unable to open database file
```

<details>
  <summary>Full error log</summary>
  
```
 14:32:42.102 | ERROR   | Task run '_get_source_filepaths-0' - Encountered exception during execution:
Traceback (most recent call last):
  File "/Users/waygr/miniforge3/envs/jump_sc/lib/python3.10/site-packages/prefect/engine.py", line 1551, in orchestrate_task_run
    result = await call.aresult()
  File "/Users/waygr/miniforge3/envs/jump_sc/lib/python3.10/site-packages/prefect/_internal/concurrency/calls.py", line 181, in aresult
    return await asyncio.wrap_future(self.future)
  File "/Users/waygr/miniforge3/envs/jump_sc/lib/python3.10/site-packages/prefect/_internal/concurrency/calls.py", line 194, in _run_sync
    result = self.fn(*self.args, **self.kwargs)
  File "/Users/waygr/miniforge3/envs/jump_sc/lib/python3.10/site-packages/cytotable/sources.py", line 70, in _get_source_filepaths
    for table_name in _duckdb_with_sqlite()
duckdb.InvalidInputException: Invalid Input Error: Attempting to execute an unsuccessful or closed pending query result
Error: Invalid Error: unable to open database file

14:32:42.118 | ERROR   | Task run '_get_source_filepaths-0' - Finished in state Failed('Task run encountered an exception: duckdb.InvalidInputException: Invalid Input Error: Attempting to execute an unsuccessful or closed pending query result\nError: Invalid Error: unable to open database file\n')

14:32:42.119 | ERROR   | Flow run 'berserk-gopher' - Encountered exception during execution:
Traceback (most recent call last):
  File "/Users/waygr/miniforge3/envs/jump_sc/lib/python3.10/site-packages/prefect/engine.py", line 669, in orchestrate_flow_run
    result = await flow_call.aresult()
  File "/Users/waygr/miniforge3/envs/jump_sc/lib/python3.10/site-packages/prefect/_internal/concurrency/calls.py", line 181, in aresult
    return await asyncio.wrap_future(self.future)
[...]

PR #45, opened 2023-04-18
Enable SQLite data to be read from cloud-based sources

d33bs:
# Description

This PR addresses #43 by enabling SQLite data to be read from cloud-based locations. This work involves caching cloud-based SQLite data locally so they may be queried locally. Please see the documentation updates for more details on the exact mechanism and examples.

Note, while addressing changes, I also made minor updates which are tangential but help with the performance or understanding of the changes:
- Updated documentation to exclude the use of `merge` and instead prefer `join` (leaning towards SQL-based rather than Pandas-based data linking language).
- Marked a single dask test as skipped by default. This test is memory intensive and I believe is the cause of Github Actions testing resource issues. Leaving the test will allow for further testing of this capability if desired and hopefully avoid challenges with automated testing in Github. In skipping the test I also hope to make more room for the additional testing performed to validate the new functionality added by this PR.

- [x] This change requires a documentation update.

# Checklist

### DIFF
--- a/cytotable/sources.py
+++ b/cytotable/sources.py
@@ -66,6 +66,10 @@ def _get_source_filepaths(
     """
 
     if path.is_file() and path.suffix == ".sqlite":
+        if any(f"{cloudtype}://" in str(path) for cloudtype in ["s3", "gcp", "az"]):
+            path.read_bytes()
+            path = AnyPath(path.fspath)
+
         return {
             f"{table_name}.sqlite": [{"table_name": table_name, "source_path": path}]
             for table_name in _duckdb_with_sqlite()
--- a/cytotable/sources.py
+++ b/cytotable/sources.py
@@ -10,7 +10,7 @@ from cloudpathlib import AnyPath, CloudPath
 from prefect import flow, task
 
 from cytotable.exceptions import DatatypeException, NoInputDataException
-from cytotable.utils import _duckdb_with_sqlite, _cache_cloudpath_to_local
+from cytotable.utils import _cache_cloudpath_to_local, _duckdb_with_sqlite
 
 
 @task
@@ -33,7 +33,7 @@ def _build_path(
     """
 
     # form a path using cloudpathlib AnyPath, stripping certain characters
-    processed_path = AnyPath(path.strip("'\" "))
+    processed_path = AnyPath(str(path).strip("'\" "))
 
     # set the client for a CloudPath
     if isinstance(processed_path, CloudPath):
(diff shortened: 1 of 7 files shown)


cytotable/sources.py
```diff
@@ -45,64 +46,100 @@ def _build_path(
...
     targets: List[str],
-    source_datatype: Optional[str] = None,
 ) -> Dict[str, List[Dict[str, Any]]]:
     """
     Gather dataset of filepaths from a provided directory path.
 
     Args:
         path: Union[pathlib.Path, Any]:
-            Path to seek filepaths within.
+            Path as directory to seek filepaths within or file reference.
```
gwaybio, line 55:
```suggestion
            Either a directory path to seek filepaths within or a path directly to a file.
```

Consider also specifying something to the extend that the Path can be local or on the cloud.

gwaybio: suggestion just to boost documentation clarity

cytotable/utils.py
```diff
@@ -75,3 +79,38 @@ def _duckdb_with_sqlite() -> duckdb.DuckDBPyConnection:
...
+            for cloudtype in ["s3://", "gc://", "az://"]
+        )
+        # check that the path is a file (caching won't work with a dir)
+        and AnyPath(path).is_file()
+        # check that the file is of sqlite type
+        # (other file types will be handled remotely in cloud)
+        and AnyPath(path).suffix.lower() == ".sqlite"
+    ):
+        # incur a read which will trigger caching of the file
+        CloudPath(path).read_bytes()
```
gwaybio, line 111: Do you have a sense of how long this will take? I presume this will be slow

gwaybio (approved):
LGTM!

We should wait for @falquaddoomi 's review to handle more of the technical elements

cytotable/utils.py
```diff
@@ -75,3 +79,38 @@ def _duckdb_with_sqlite() -> duckdb.DuckDBPyConnection:
...
+        # update the path to be the local filepath for reference in CytoTable ops
+        path = CloudPath(path).fspath
+
+    # cast the result as a pathlib.Path
+    return pathlib.Path(path)
```
falquaddoomi, line 116: From the docs, seems that just referencing `.fspath` is enough to cause the remote file to be cached, which I imagine would make it unnecessary to call `read_bytes()` on it: https://cloudpathlib.drivendata.org/stable/caching/#accessing-the-cached-version-directly-read-only.

Also, I could imagine `clouthpathlib` being extended with additional providers in the future. If that seems like something you'd want, you might defer to the library to determine if it supports the given URL, e.g.:

```python
candidate_path = AnyPath(path)
  
if candidate_path.is_file() and candidate_path.suffix.lower() == ".sqlite":
  try:
    path = CloudPath(path).fspath
  except clouthpathlib.exceptions.CloudPathException:
    # you might consider logging here, up to you
    pass
    
return pathlib.Path(path)
```

falquaddoomi (commented):
Nice work; I left a few comments.

cytotable/utils.py
d33bs: This will vary based on where CytoTable is running (network access, hardware data storage type), the cloud provider (AWS, GCP, etc.), and how the bucket is configured (geolocation, throttling, etc). For AWS S3 locations [`cloudpathlib` leverages `boto3` via the `S3Client`](https://github.com/drivendataorg/cloudpathlib/blob/master/cloudpathlib/s3/s3client.py#L153) to perform downloads.

Wondering aloud: would it be a good idea to provide a caveat about larger cloud-based SQLite data (that performance may vary)?

cytotable/sources.py
d33bs: So much better, thanks!

cytotable/utils.py
d33bs: Thanks for this, great thoughts! I wrestled with both of these a bit and welcome the chance to improve.
- `fspath` vs `read_bytes`: great point, I'll get this changed. I don't feel strongly about it but wondering aloud if `read_bytes` is less presumptive about the underlying operations (perhaps a cloudpath could be understood to be a type of "filesystem" in some circumstances).
- additional providers: One reason for the explicit checks was that I couldn't get [`CloudPath.is_valid_cloudpath`](https://cloudpathlib.drivendata.org/v0.6/api-reference/cloudpath/#cloudpathlib.cloudpath.CloudPath.is_valid_cloudpath) to work in expected ways. I like the alternative of a try/except here better for the reasons you mentioned. I'll these a try in an update.

cytotable/sources.py
```diff
@@ -180,7 +217,7 @@ def _filter_source_filepaths(
             # ensure the filesize is greater than 0
             if file["source_path"].stat().st_size > 0
             # ensure the datatype matches the source datatype
-            and file["source_path"].suffix == f".{source_datatype}"
+            if file["source_path"].suffix == f".{source_datatype}"
```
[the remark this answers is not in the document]
d33bs: Thanks! This was accidental during the course of development, mostly with regards to how cloud directories vs cloud files could be handled within `cytotable`. I like the `and` better for readability and will move back to this.

cytotable/sources.py
```diff
@@ -45,64 +46,100 @@ def _build_path(
...
+                or str(file.stem).lower() in [target.lower() for target in targets]
+                # checks for sqlite extension (which may include compartment + metadata names)
+                or file.suffix.lower() == ".sqlite"
             )
-            .arrow()["table_name"]
-            .to_pylist()
-            if any(target.lower() in table_name.lower() for target in targets)
-        }
+        ]
+    )
```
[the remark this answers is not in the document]
d33bs: Thanks, this is great! I'll move over to this new format as you suggested; the edge case of an empty list is unintentional (maybe at some point property-based / hypothesis testing in this project could capture things like this!).

falquaddoomi (approved):

PR #46, opened 2023-04-18
Adding Legacy Presets for CytoTable 

axiomcura:
_modified from [EmbeddedArtistry](https://embeddedartistry.com/blog/2017/08/04/a-github-pull-request-template-for-your-projects/)_
_referenced with modifications from [pycytominer](https://github.com/cytomining/pycytominer/blob/master/.github/PULL_REQUEST_TEMPLATE.md)_

# Description

This PR focuses adding `CellProfiler` legacy presets for `CytoTable`'s `convert()` function. 

**What motivated you to make this change?**

[Datasets](https://nih.figshare.com/articles/dataset/Cell_Health_-_Cell_Painting_Single_Cell_Profiles/9995672/1) produced by older version of [CellProfiler](https://github.com/CellProfiler/CellProfiler) (For example: ver 2.X) contain different naming schemes, hence causing errors with the current presets that exists within `CytoTable` 

The purpose of this PR is to introduce a new preset named `cellprofiler_legacy`. This preset includes the necessary naming and merging configurations for converting datasets generated from previous versions of CellProfiler (2.X) into parquet files.

- [X] Enhancement (adds functionality).

# Checklist

- [X] I have read the [CONTRIBUTING.md](CONTRIBUTING.md) guidelines.
- [X] My code follows the style guidelines of this project.
- [X] I have performed a self-review of my own code.
- [X] I have commented my code, particularly in hard-to-understand areas.
- [X] My changes generate no new warnings.
- [ ] New and existing unit tests pass locally with my changes.
- [ ] I have deleted all non-relevant text in this pull request template.

### DIFF
--- a/cytotable/presets.py
+++ b/cytotable/presets.py
@@ -153,6 +153,59 @@ config = {
                 AND per_nuclei.Metadata_Nuclei_Number_Object_Number = per_cytoplasm.Metadata_Cytoplasm_Parent_Nuclei
             """,
     },
+    "cellprofiler_legacy": {
+        # version specifications using related references
+        "CONFIG_SOURCE_VERSION": {
+            "cellprofiler": "v2.X",
+            "pycytominer": "c90438fd7c11ad8b1689c21db16dab1a5280de6c",
+        },
+        # names of source table compartments (for ex. cells.csv, etc.)
+        "CONFIG_NAMES_COMPARTMENTS": ("cells", "nuclei", "cytoplasm"),
+        # names of source table metadata (for ex. image.csv, etc.)
+        "CONFIG_NAMES_METADATA": ("image",),
+        # column names in any compartment or metadata tables which contain
+        # unique names to avoid renaming
+        "CONFIG_IDENTIFYING_COLUMNS": (
+            "ImageNumber",
+            "Metadata_Well",
+            "Parent_Cells",
+            "Parent_Nuclei",
+            "Cytoplasm_Parent_Cells",
+            "Cytoplasm_Parent_Nuclei",
+            "Cells_ObjectNumber",
+            "Nuclei_ObjectNumber",
+        ),
+        # chunk size to use for join operations to help with possible performance issues
+        # note: this number is an estimate and is may need changes contingent on data
+        # and system used by this library.
+        "CONFIG_CHUNK_SIZE": 1000,
+        # chunking columns to use along with chunk size for join operations
+        "CONFIG_CHUNK_COLUMNS": ("Metadata_ImageNumber",),
+        # compartment and metadata joins performed using DuckDB SQL
+        # and modified at runtime as needed
+        "CONFIG_JOINS": """
+            WITH Image_Filtered AS (
+                SELECT
+                    Metadata_ImageNumber,
+                    Image_Metadata_Well,
+                    Image_Metadata_Plate
+                FROM
+                    read_parquet('image.parquet')
+                )
+            SELECT
+                *
+            FROM
+                Image_Filtered AS image
+            LEFT JOIN read_parquet('cytoplasm.parquet') AS cytoplasm ON
+                cytoplasm.Metadata_ImageNumber = image.Metadata_ImageNumber
+            LEFT JOIN read_parquet('cells.parquet') AS cells ON
+                cells.Metadata_ImageNumber = cytoplasm.Metadata_ImageNumber
+                AND cells.Cells_ObjectNumber = cytoplasm.Metadata_Cytoplasm_Parent_Cells
+            LEFT JOIN read_parquet('nuclei.parquet') AS nuclei ON
+                nuclei.Metadata_ImageNumber = cytoplasm.Metadata_ImageNumber
+                AND nuclei.Nuclei_ObjectNumber = cytoplasm.Metadata_Cytoplasm_Parent_Nuclei
+        """
+    }
 }
 """
 Configuration presets for pycytominer-transform
--- a/cytotable/presets.py
+++ b/cytotable/presets.py
@@ -204,8 +204,8 @@ config = {
             LEFT JOIN read_parquet('nuclei.parquet') AS nuclei ON
                 nuclei.Metadata_ImageNumber = cytoplasm.Metadata_ImageNumber
                 AND nuclei.Nuclei_ObjectNumber = cytoplasm.Metadata_Cytoplasm_Parent_Nuclei
-        """
-    }
+        """,
+    },
 }
 """
 Configuration presets for pycytominer-transform


cytotable/presets.py
```diff
@@ -153,6 +153,59 @@
                 AND per_nuclei.Metadata_Nuclei_Number_Object_Number = per_cytoplasm.Metadata_Cytoplasm_Parent_Nuclei
             """,
     },
+    "cellprofiler_legacy": {
+        # version specifications using related references
+        "CONFIG_SOURCE_VERSION": {
+            "cellprofiler": "v2.X",
```
d33bs, line 159: Looking at the nam of this preset and the version here, would it make sense to specify a version range for CellProfiler? For example, `v2.x-v3.x`, or something like this?

d33bs (approved):
Nice additions! I added a few comments and felt that overall this LGTM!

When you have the chance, could I ask that you update your branch to the latest code from CytoTable? There are recent updates which may allow the Github Actions tests to successfully run (the current failure doesn't happen due to the changes here).

One additional thought that I wanted to pose: would it make sense to add a small test for this? This could prove useful for demonstrating what is expected in the results.

gwaybio (approved):

cytotable/presets.py
gwaybio: > When did the renaming of the column started to happen? 

I do not know

jenna-tomkinson:
@gwaybio @d33bs @axiomcura 

After talking to Erik, I think I have found something important that would greatly impact this PR. 

This PR is meant to help with different table names due to different CellProfiler versions. The idea was that from when Cell-Health was run with version 2.2.0 ([see documentation here for the CellProfiler pipelines](https://github.com/broadinstitute/cell-health/tree/master/0.download-data/cellprofiler_pipelines)) to the newest version, the naming convention for the tables in `ExportToDatabase` has changed.

**I have found evidence that this is incorrect, and the naming convention has not changed between versions!**

To start, we can definitely see that in the [Cell-Health data SQLite files](https://nih.figshare.com/articles/dataset/Cell_Health_-_Cell_Painting_Single_Cell_Profiles/9995672), the tables are not named with the `Per_*` convention.

image

But, if you go to the commit for the v2.2.0 of CellProfiler and [look at the documentation in the `ExportToDatabase` module](https://github.com/CellProfiler/CellProfiler/blob/ff776fe0121b979a5f4e9546f249f745674e0a67/cellprofiler/modules/exporttodatabase.py), it specifically states:

```
The database is set up with two primary tables. These tables are the
<i>Per_Image</i> table and the <i>Per_Object</i> table (which may have a prefix if you
specify)
```

That means that the SQLite files outputted by CellProfiler using [the `analysis.cppipe` file](https://github.com/broadinstitute/cell-health/blob/30ea5de393eb9cfc10b575582aa9f0f857b44c59/0.download-data/cellprofiler_pipelines/analysis.cppipe#L456), the tables should be named with the convention `Per_`.

This indicates to me that likely the files that were outputted by `ExportToSpreadsheet` were used to create a database which I think are the ones that are uploaded to figshare. 

@gwaybio I might not be remembering correctly, but did we have a passing conversation that the Broad used a separate function to create the database from the csv files exported from ExportToSpreadsheet? This is one reason I can come up with for why there is this difference. The other is that the databases were edited post-analysis run.

I hope that this helps! Please let me know if I need to clarify anything!

bethac07:
To add a bit of context - in CellProfiler, if you're exporting database tables (or spreadsheets) it is optional to either create single tables/sheets per object, or make a "per_object" table/sheet. That was true in 2 and still in 4 (and even though we're likely going to be massively rewriting export in 5, likely in 5 and forward too)

axiomcura:
This is great information! Thank you @bethac07 and @jenna-tomkinson for proving context. This is making more sense :smile:

jenna-tomkinson:
@bethac07 Thank you for this comment! I appreciate your input on this!

I think where we are getting confused is with the [Cell-Health project](https://github.com/broadinstitute/cell-health), when looking at the [SQLite files on figshare](https://nih.figshare.com/articles/dataset/Cell_Health_-_Cell_Painting_Single_Cell_Profiles/9995672), we are not seeing the "Per_Object" naming convention (see previous comment).

I double-checked the pipeline on GitHub and saw that the [parameter used in the `ExportToDatabase` module](https://github.com/broadinstitute/cell-health/blob/30ea5de393eb9cfc10b575582aa9f0f857b44c59/0.download-data/cellprofiler_pipelines/analysis.cppipe#L477) was "Single Object View". 

When I attempted to use "Single Object View" in [my pipeline](https://github.com/WayScience/nf1_cellpainting_data/blob/main/2.cellprofiler_analysis/NF1_analysis_plate1_plate2.cppipe) using v4.2.5, I got an SQLite file using the "Per_Object" naming convention (see below).

image

Based on this, is there a different parameter other than "Single Object View" in v2.2.0 that causes the Cell-Health databases seen on figshare not to have the "Per_" naming convention? Or is this a version issue?

bethac07:
Not a version issue! ExportToDatabase is not being used in the pipeline you linked (`enabled:False`), the ExportToSpreadsheet module that's below it is what's making that output, so it was presumably then reassembled with cytominer-database or some script.

jenna-tomkinson:
@bethac07 Okay thank you so much! 😄 I felt like I was going crazy there!

I never even realized that `enabled` was something located in the file, maybe because I always put a bunch of notes in the GUI so the info became hidden for me in the exported cppipe file. 

I will keep this in mind for the future, and I feel more confident interpreting the cppipe files!

@axiomcura Would this be now instead of `legacy presets`, be `cytominer-database presets`?

axiomcura:
@jenna-tomkinson Yeah definitely, I was thinking more `cytominer_sqlite` since we're dealing with `sqlite` files generated from `cytominer`. 

@d33bs @gwaybio let me know if that works

Thank you all!

cytotable/presets.py
d33bs: I'm also not sure about the name changes per versions. I brought this up just in case you knew specific versions this preset would apply to beyond just 2.x. If this is uncertain, no changes necessary 🙂.

cytotable/presets.py
```diff
@@ -153,6 +153,58 @@
                 AND per_nuclei.Metadata_Nuclei_Number_Object_Number = per_cytoplasm.Metadata_Cytoplasm_Parent_Nuclei
             """,
     },
+    "cellprofiler_legacy": {
```
d33bs, line 156: I think a name change here is now necessary based on the information we have. Maybe this could be as specific as outlining `cell-health-cellprofiler-to-cyotominer-database`? Calling out the provenance of the data as it moved through the pipelines I feel could be important for later reference.

d33bs (dismissed):
Cheers on the fantastic investigation and discussion @axiomcura, @jenna-tomkinson, and @bethac07 ! Based on the information you found, I'd like to request changes to at least the name of the new preset and also the `CONFIG_SOURCE_VERSION` elements.

I'd also wonder how much of the data provenance is preserved and if it would be appropriate to cite this information within the `CITATION.cff` file (or elsewhere as appropriate) to help indicate which versions and what code was used to produce the data.

The most thorough way to verify this (if we don't know what exactly was used) might involve running the source data through CellProfiler with the pipeline file which was used for Cell-Health, using cytominer-database to create the resulting SQLite database, then validating replication between the result and what's on Figshare. I don't have a sense for the availability of the data to accomplish this, or how time consuming it may be to do, so would be interested in everyone's thoughts on what's reasonable here (this could be challenging code archeology or hardware resource intensive!).

Addressing your question @axiomcura:
> Yes definitely! I will add a small test. I think you've mentioned a small row-based subset of the data, correct?

I think a subset of the data you're using with this preset would work. I recommend writing code which may live alongside the test data which represents how the original source is modified (the data modification code wouldn't in my opinion need to be executed as part of the test suite).

gwaybio:
Is this ready to merge @d33bs ?

d33bs:
Hi @gwaybio , this is on it's way to completion. I've made some additions to the preset and citations. Next I hope to add a small test referencing a Cell-Health dataset to verify and demonstrate the use of this preset.

tests/test_convert.py
```diff
@@ -1081,3 +1082,57 @@ def test_convert_hte_cellprofiler_csv(
 
     # clean up the parsl config for other tests
     parsl.clear()
+
+
+def test_cellhealth_preset(
```
gwaybio, line 1087: I'm not sure if you do this elsewhere (I'm sure you do, actually), but it might be nice to include an explicit join test for this preset as well.

Specifically, can you load each sqlite table individually, merge, and then compare to the convert approach? (you'll probably have to replace some `nan` values) but I think it would be good to confirm accurate join.

gwaybio, line 1087:
```suggestion
def test_legacy_cellprofiler_cytominerdatabase_cellheatlth_preset(
```

gwaybio: or maybe just verbatim: `test_cell_health_cellprofiler_to_cytominer_database_legacy`

gwaybio (approved):
LGTM! I had one optional test comment, which may be outside scope or not helpful. Please use your judgment and add the test if appropriate. No need for me to review again 🎉

tests/test_convert.py
d33bs: Great point, thanks! I'll add this in.

d33bs: Thanks for this, I'll go with the verbatim option and add this in.

cytotable/presets.py
```diff
@@ -153,6 +153,58 @@
...
+        # version specifications using related references
+        "CONFIG_SOURCE_VERSION": {
```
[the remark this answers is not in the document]
d33bs: This has been added in further changes.

cytotable/presets.py
d33bs: This has been added in further changes.

cytotable/presets.py
d33bs: Left as-is given uncertainty.

d33bs:
Note: tests appeared to momentarily fail with Python or OS specific aspects. Full column value sorting was added to legacy preset testing comparisons as a safety measure towards resolving this in the future. As testing now passes, going to merge momentarily.

d33bs:
Many thanks to @axiomcura , @jenna-tomkinson, @bethac07 , and @gwaybio for your tremendous help with things here!

Issue #38, opened 2023-03-23
Duckdb's `TypeMismatchException` raised in CytoTable's convert() workflow due to `nan` values being stored as strings instead of expected types 


axiomcura:
`CytoTable`'s `convert()` function seems to capture `nan`'s as string types within the cell-helath dataset causing `duckdb` to raise `duckdb.TypeMismatchException` error. 

Below is the code to replicate the problem:
```python
import pathlib
import cytotable

# sqlite file path
sqlite_file = str(pathlib.Path("./SQ00014613.sqlite").resolve(strict=True))

# execute covert workflow 
data = cytotable.convert(sqlite_file, dest_path="./parquet_data/test.parquet", dest_datatype="parquet", preset="cellprofiler_sqlite", source_datatype="sqlite")
```

link to download [data](https://nih.figshare.com/ndownloader/files/18506036)

### From Prefect:
```text
13:38:45.442 | ERROR   | Task run '_read_data-1' - Encountered exception during execution:
Traceback (most recent call last):
  File "/home/erikserrano/Programs/miniconda3/envs/pycytominer/lib/python3.9/site-packages/prefect/engine.py", line 1533, in orchestrate_task_run
    result = await run_sync(task.fn, *args, **kwargs)
  File "/home/erikserrano/Programs/miniconda3/envs/pycytominer/lib/python3.9/site-packages/prefect/utilities/asyncutils.py", line 156, in run_sync_in_interruptible_worker_thread
    tg.start_soon(
  File "/home/erikserrano/Programs/miniconda3/envs/pycytominer/lib/python3.9/site-packages/anyio/_backends/_asyncio.py", line 662, in __aexit__
    raise exceptions[0]
  File "/home/erikserrano/Programs/miniconda3/envs/pycytominer/lib/python3.9/site-packages/anyio/to_thread.py", line 31, in run_sync
    return await get_asynclib().run_sync_in_worker_thread(
  File "/home/erikserrano/Programs/miniconda3/envs/pycytominer/lib/python3.9/site-packages/anyio/_backends/_asyncio.py", line 937, in run_sync_in_worker_thread
    return await future
  File "/home/erikserrano/Programs/miniconda3/envs/pycytominer/lib/python3.9/site-packages/anyio/_backends/_asyncio.py", line 867, in run
    result = context.run(func, *args)
[...]

PR #50, opened 2023-05-09
Add SQLite mixed-type data handling capabilities

d33bs:
# Description

This change seeks to address #38 through dynamic handling of mixed-type SQLite column data when exceptions of this kind are encountered with DuckDB SQLite queries.

When mixed type exceptions are encountered, the design intent here strives to keep as much data as possible while assuming one or many values of mismatching type within the SQLite table being queried. SQLite3 SQL case statements are formed to convert mismatched storage class values to NULL's for every column in the table being queried. These SQLite NULL's are interpreted by `sqlite3` as Python `None`'s, which are then interpreted as `pyarrow.null()`'s for export to Parquet.

For SQLite data sources where mixed typing exceptions are not encountered, the default capabilities through DuckDB function as usual.

# Checklist


cytotable/utils.py
```diff
@@ -149,6 +149,93 @@ def _duckdb_reader() -> duckdb.DuckDBPyConnection:
...
+        source_path: str:
+            A str which is a path to a SQLite database file
+        table_name: str:
+            The name of the table being queried
+        chunk_size: int:
+            Row count to use for chunked output
+        offset: int:
+            The offset for chunking the data from source.
+        dest_path: str:
+            Path to store the output data.
```
falquaddoomi, line 174: This is a very minor nitpick, but some argument descriptions end in a period and some don't. I don't have a preference between the two, but I think it'd be better if they were consistent.

cytotable/utils.py
```diff
@@ -149,6 +149,93 @@ def _duckdb_reader() -> duckdb.DuckDBPyConnection:
...
+            FROM pragma_table_info(:table_name);
+            """,
+            {"table_name": table_name},
+        )
+
+        # gather column metadata details as list of dictionaries
+        column_info = [
+            dict(zip([desc[0] for desc in cursor.description], row))
+            for row in cursor.fetchall()
+        ]
```
falquaddoomi, line 204: Nifty! 🎉

tests/test_convert.py
```diff
@@ -823,6 +828,62 @@ def test_convert_cellprofiler_sqlite_pycytominer_merge(
...
+        # run a more nuanced query through sqlite
+        # to handle the mixed types
+        if "Mismatch Type Error" in str(duckdb_exc):
+            result = _sqlite_mixed_type_query_to_parquet(
+                source_path=example_sqlite_mixed_types_database,
+                table_name=table_name,
+                chunk_size=2,
+                offset=0,
+                result_filepath=result_filepath,
+            )
```
falquaddoomi, line 862: Same comment here as before regarding using `duckdb.TypeMismatchException` over its ancestor `duckdb.Error`.

tests/conftest.py
```diff
@@ -488,3 +489,58 @@ def example_s3_endpoint(
...
+                "INSERT INTO tbl_a (col_integer, col_text, col_blob, col_real)"
+                "VALUES (?, ?, ?, ?);"
+            ),
+            err_values,
+        )
+
+    yield filepath
+
+    # after completing the tests, remove the file
+    pathlib.Path(filepath).unlink()
```
falquaddoomi, line 546: Cool that pytest uses generators to allow for cleanup steps after creating a fixture. I don't know if this is necessary since I'm not super familiar with pytest, but this might be more robust as a try/finally, e.g.:

```suggestion
    try:
        yield filepath
    finally:
        # after completing the tests, remove the file
        pathlib.Path(filepath).unlink()
```

That way, even if the coprocess to which you're yielding never returns (not that that's likely, but who knows?), you're guaranteed to delete the file. Again, this might be too cautious, so I'll leave it up to you to choose. (On a side note, I have no idea why github's syntax colorer isn't coloring "finally", but that's definitely valid Python code...)

cytotable/convert.py
```diff
@@ -152,17 +153,36 @@ def _source_chunk_to_parquet(
...
+        _duckdb_reader().execute(
+            f"""
+            COPY (
+                {base_query}
+                LIMIT {chunk_size} OFFSET {offset}
+            ) TO '{result_filepath}'
+            (FORMAT PARQUET);
+            """
+        )
+    except duckdb.Error as e:
```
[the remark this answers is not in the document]
d33bs: Thanks for calling this forward! I noticed mixed behavior with DuckDB Python Exceptions which resulted in the approach here. I'll give the more specific exception another attempt here (very much prefer it! 🙂 ).

cytotable/utils.py
d33bs: Good point, thanks, I'll update these!

cytotable/convert.py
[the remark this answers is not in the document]
d33bs: Digging into this more I found that only generic errors are raised from the context of when this happens from the `sqlite_scanner` extension. I'm going to raise an issue for that extension with the hope that we may get more feedback here. In the meantime, I'll plan to leave things as-is.

See the following demonstration code:

```python
import sqlite3

import duckdb

# create a temporary sqlite connection
filepath = ".example.sqlite"

# statements for creating database with simple structure
create_stmts = """
    DROP TABLE IF EXISTS tbl_a;
    CREATE TABLE tbl_a (
        col_integer INTEGER NOT NULL
        ,col_text TEXT
        ,col_blob BLOB
        ,col_real REAL
    );
    """

# some example values to insert into the database
insert_vals = ["nan", "sample", b"sample_blob", 0.5]

# create the database and insert some data into it
with sqlite3.connect(filepath) as connection:
    connection.executescript(create_stmts)

    connection.execute(
        (
            "INSERT INTO tbl_a (col_integer, col_text, col_blob, col_real)"
            "VALUES (?, ?, ?, ?);"
        ),
        insert_vals,
    )

try:
    duckdb.connect().execute(
        f"""
    /* install and load sqlite plugin for duckdb */
    INSTALL sqlite_scanner;
    LOAD sqlite_scanner;

    /* perform query on tbl_a table */
    SELECT * FROM sqlite_scan('{filepath}', 'tbl_a');
    """
    ).arrow()
    
except duckdb.TypeMismatchException as type_mismatch_e:
    print("Found a type mismatch exception: ", type_mismatch_e)
    
except Exception as e:
    print("Found a different error of type: ", type(e), e, sep="\n")
```

tests/test_convert.py
d33bs: Similar to the previous comment, leaving this as-is for right now.

cytotable/convert.py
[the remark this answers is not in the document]
falquaddoomi: Nice detective work, and a pity the extension isn't throwing the correct exception. Sounds good regarding leaving things as-is for now, since there aren't any better options.

gwaybio:
@falquaddoomi @d33bs - is this ready to merge? Will it solve #38 (@jenna-tomkinson and @MattsonCam are currently blocked by it for the JUMP single cell analysis)

gwaybio:
Update, @d33bs - just saw your comment in https://github.com/cytomining/CytoTable/issues/38#issuecomment-1625615593

gwaybio (approved):
LGTM! Excited to get this merged in.

Issue #58, opened 2023-05-19
Enhance convert() default or unspecified parameters with preferential choices


d33bs:
This issue highlights the potential to enable "smarter" default or unspecified parameter decisions for `convert()` such that parameters match, for example, the storage type or align themselves in unsurprising ways. For example, if a SQLite file is provided, default parameters could align to patterns found for those data sources in order to create a better user experience. Currently, many of the defaults stem from CellProfiler CSV's, which sometimes causes unexpected challenges. Inspired by #57  .

PR #65, opened 2023-06-01
Enable preset overrides

### LINKED ISSUE #64: Enable convert preset parameter overrides
This issue highlights the need to allow parameters to override presets when needed. For instance, if a preset configuration is used and only some of the values are needed from the preset, and others are desired to be overridden, a user should be able to specify those parameters when used by convert. The current functionality appears to use only the preset parameters when they are specified. 

Based on conversations with @gwaybio.
References #61

d33bs:
# Description

This PR makes changes to allow for preset overrides from `convert` parameters. Additionally, a test appeared to require sorting as a result of this change so a minor change was added for this.

Closes #64 

# Checklist

### DIFF
--- a/cytotable/convert.py
+++ b/cytotable/convert.py
@@ -1004,7 +1004,7 @@ def convert(  # pylint: disable=too-many-arguments,too-many-locals
     chunk_size: Optional[int] = None,
     infer_common_schema: bool = True,
     drop_null: bool = True,
-    preset: Optional[str] = None,
+    preset: Optional[str] = "cellprofiler_csv",
     parsl_config: Optional[parsl.Config] = None,
     **kwargs,
 ) -> Union[Dict[str, List[Dict[str, Any]]], str]:
--- a/cytotable/convert.py
+++ b/cytotable/convert.py
@@ -1029,7 +1029,7 @@ def convert(  # pylint: disable=too-many-arguments,too-many-locals
         metadata: Union[List[str], Tuple[str, ...]]:
             Metadata names to use for conversion.
         compartments: Union[List[str], Tuple[str, str, str, str]]:
-            (Default value = DEFAULT_COMPARTMENTS)
+            (Default value = None)
             Compartment names to use for conversion.
         identifying_columns: Union[List[str], Tuple[str, ...]]:
             Column names which are used as ID's and as a result need to be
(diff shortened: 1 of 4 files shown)


cytotable/convert.py
```diff
@@ -994,27 +994,17 @@ def convert(  # pylint: disable=too-many-arguments,too-many-locals
...
-    chunk_size: Optional[int] = cast(
-        int, config["cellprofiler_csv"]["CONFIG_CHUNK_SIZE"]
-    ),
+    joins: Optional[str] = None,
+    chunk_columns: Optional[Union[List[str], Tuple[str, ...]]] = None,
+    chunk_size: Optional[int] = None,
     infer_common_schema: bool = True,
     drop_null: bool = True,
-    preset: Optional[str] = None,
+    preset: Optional[str] = "cellprofiler_csv",
```
falquaddoomi, line 1007: I'd definitely call out clearly in the docs that this has been changed from `None` to a default preset. At least be sure to update the docstring to indicate that this is the default value for `preset`; as of now it still says `None` is the default.

falquaddoomi (commented):
This seems like a reasonable change. Aside from my comment, I don't see any reason not to approve it.

Outside the scope of this PR, but a few thoughts for later: I don't know how common it will be to add presets, but perhaps if it does become common you might consider a "plugin" system where presets are stored as JSON or some other user-editable format (vs. having to edit a dictionary in the source), then made available as arguments to this function. You could also factor out the application of a preset into a seperate function, as applying presets is starting to take up more than a few lines in this core function and I can only imagine it taking up more as you add options.

falquaddoomi (approved):

cytotable/convert.py
d33bs: Thank you! I've updated the documentation to reflect this. While this does change the preset default, the individual preset parameters were already coming from this preset (set from the parameter defaults as opposed to the preset logic within). My hope is that we can reach #58 at some point to avoid the use of preset options when one is not provided and instead dynamically choose options which make sense based on the data source(s).

d33bs:
Thanks @falquaddoomi for the review and comments! I love the idea of being able to supply JSON or similar as preset options to CytoTable and have created #66 to address this in the future. Please note: I also updated tests to avoid the use of Dagger, which has been showing unpredictable results concerning temporary directories within Github Actions images.

PR #78, opened 2023-07-26
Change drop_nulls default in `convert()` to false

### LINKED ISSUE #56: Default for `drop_nulls=True` causes loss of data
When reviewing the number of segmented single cells per image, I noticed a big discrepancy between the output from CytoTable and CellProfiler.

After looking at the CytoTable `convert` function, the parameter `drop_nulls` is defaulted to true, which means that a significant amount of data (rows) is removed due to having at least one column with an NaN value. 

In my original example, I run through a dictionary with paths for each plate to run `convert`:

```python
# run through each run with each set of paths based on dictionary
for plate, info in plate_info_dictionary.items():
    source_path = info["source_path"]
    dest_path = info["dest_path"]
    print(f"Performing merge single cells and conversion on {plate}!")

    # merge single cells and output as parquet file
    convert(
        source_path=source_path,
        dest_path=dest_path,
        dest_datatype=dest_datatype,
        preset=preset,
    )
    print(f"Merged and converted {pathlib.Path(dest_path).name}!")

    # add single cell count per well as metadata column to parquet file and save back to same path
    sc_utils.add_sc_count_metadata_file(
        data_path=dest_path, well_column_name="Image_Metadata_Well", file_type="parquet"
    )
    print(f"Added single cell count as metadata to {pathlib.Path(dest_path).name}!")
```
With the default parameter `drop_nulls` set to true, I get this output for the last variable run in this for loop:

image

After adding the `drop_nulls` parameter to the above function and setting it to False, we get this output for the same plate:

image

Since we preprocess the features before analysis, we can drop columns containing NaN values without removing the segmented single cells entirely. This will be an easy fix to default to False instead and let the user decide if they are okay with removing single cells/rows.

gwaybio:
# Description

Changes the `convert()` default to not drop nulls by default. This is more in-line with expected behavior.

Closes #56

# Checklist

### DIFF
--- a/cytotable/convert.py
+++ b/cytotable/convert.py
@@ -1180,7 +1180,7 @@ def convert(  # pylint: disable=too-many-arguments,too-many-locals
     chunk_columns: Optional[Union[List[str], Tuple[str, ...]]] = None,
     chunk_size: Optional[int] = None,
     infer_common_schema: bool = True,
-    drop_null: bool = True,
+    drop_null: bool = False,
     data_type_cast_map: Optional[Dict[str, str]] = None,
     preset: Optional[str] = "cellprofiler_csv",
     parsl_config: Optional[parsl.Config] = None,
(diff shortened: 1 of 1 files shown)


d33bs (approved):
Cheers, LGTM! Marking this as approved with the understanding that the tests will very likely pass after updating from main. The test failure is I believe unrelated to the changes here and instead pertains to issues which are fixed from the merge of #79.

gwaybio:
Thanks @d33bs - tests pass now, I'm going to merge!

gwaybio:
Tests passed here, but not in the merge: https://github.com/cytomining/CytoTable/actions/runs/5696050273/job/15440405650 🤔

d33bs:
Going to follow up on the errors to investigate further.

PR #82, opened 2023-08-01
Add SQLite mixed-type compatibility for _get_table_columns_and_types

d33bs:
# Description

This PR adds changes which seek to further address #38. This extends the work found in #50 for greater compatibility with mixed-types in `_get_table_columns_and_types`.

## What is the nature of your change?

- [x] Bug fix (fixes an issue).
- [ ] Enhancement (adds functionality).
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected).
- [ ] This change requires a documentation update.

# Checklist

Please ensure that all boxes are checked before indicating that a pull request is ready for review.

- [x] I have read the [CONTRIBUTING.md](CONTRIBUTING.md) guidelines.
- [x] My code follows the style guidelines of this project.
- [x] I have performed a self-review of my own code.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have made corresponding changes to the documentation.
- [x] My changes generate no new warnings.
- [x] New and existing unit tests pass locally with my changes.
- [x] I have added tests that prove my fix is effective or that my feature works.
- [x] I have deleted all non-relevant text in this pull request template.

Issue #81, opened 2023-07-31
Resolve testing issue with missing dependent workflow data files


d33bs:
This issue focuses on a bug which periodically shows up from the test `test_convert_s3_path_sqlite` where single data extraction chunks appear to not be available for subsequent dependent workflow tasks. Examples of this occurring: see [here](https://github.com/cytomining/CytoTable/actions/runs/5682733246/job/15401772592?pr=46), [here](https://github.com/cytomining/CytoTable/actions/runs/5674349561/job/15377707836?pr=78), and [here](https://github.com/cytomining/CytoTable/actions/runs/5696050273/job/15440405650).

Patterns:
- Often occurs within GitHub Actions runner images for [`ubuntu-latest`](https://github.com/actions/runner-images#available-images)(`ubuntu-22.04` as of the time of this writing).
- Often does not occur during local development test runs on MacOS (I've personally been unable to witness this occurring in a local development environment).
- Often occurs with tests in Python 3.8 (Python >= 3.9, < 3.11 appear to succeed when they reach adjacent completion)
- `FileNotFoundError` is emitted for extracted chunks of data in no apparent pattern (the files are often sourced from different chunk offsets or tables).
- In the context of `test_convert_s3_path_sqlite`, `cytotable.convert` is used 3 times. The first two `convert`'s succeed without failure, the third is typically the one which fails when this error occurs (error seems to manifest on running the same process multiple times with the first runs succeeding).
- Pytest fixture nesting related to temporary directory usage in the testing suite does not appear to be a root cause (a hotfix was attempted [here](https://github.com/cytomining/CytoTable/pull/79) with issues persisting after merge [here](https://github.com/cytomining/CytoTable/actions/runs/5696050273/job/15440405650)).
- Adding simulated `time.sleep` delays for `_source_chunk_to_parquet` does not seem to exhibit the issue consistently (possibly indicating workflows steps are completing as designed).
- Both [DuckDB SQL `COPY`](https://duckdb.org/docs/sql/statements/copy.html) and [pyarrow.parquet.write_table](https://arrow.apache.org/docs/python/generated/pyarrow.parquet.write_table.html) exhibit the issue (moving to one or the other doesn't resolve the issue).

d33bs:
Marking this as closed with merge of #83 , deferring to open new issues if this occurs again in the same way.

PR #79, opened 2023-07-27
Fix test_convert_s3_path_sqlite test failure inconsistencies

d33bs:
# Description

The changes in this PR help address test failures found in recent automated test job runs (see [here](https://github.com/cytomining/CytoTable/actions/runs/5682733246/job/15401772592?pr=46), and [here](https://github.com/cytomining/CytoTable/actions/runs/5674349561/job/15377707836?pr=78)). These appear to have been occurring due to temporary directory creation (and related destruction post yield) race conditions with nested pytest fixture usage. Failures are somewhat inconsistent; local and GitHub action based testing sometimes succeed and other times fail. To address this, manual temporary directory creation for `test_convert_s3_path_sqlite` was added to avoid the destination path being removed prior to workflows completing.

Thank you for your review and any thoughts you may have towards changes!

## What is the nature of your change?

- [x] Bug fix (fixes an issue).
- [ ] Enhancement (adds functionality).
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected).
- [ ] This change requires a documentation update.

# Checklist

Please ensure that all boxes are checked before indicating that a pull request is ready for review.

- [x] I have read the [CONTRIBUTING.md](CONTRIBUTING.md) guidelines.
- [x] My code follows the style guidelines of this project.
- [x] I have performed a self-review of my own code.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have made corresponding changes to the documentation.
- [x] My changes generate no new warnings.
- [x] New and existing unit tests pass locally with my changes.
- [x] I have added tests that prove my fix is effective or that my feature works.
- [x] I have deleted all non-relevant text in this pull request template.

PR #83, opened 2023-08-01
Resolve test_convert_s3_path_sqlite test failures with explicit local_cache_dir

d33bs:
# Description

This PR seeks to address #81 where tests periodically fail with `test_convert_s3_path_sqlite` (see issue for greater specificity on the patterns associated). Initially I made the mistake of believing this was caused due to nested Pytest fixture behavior with temporary directories supplied to tests (see PR #79). This PR reverts those changes and adds different ones to prevent the same tests failures from persisting.

I believe this test is failing due to [Cloudpathlib's default temporary cache behavior (`"tmp_dir"`)](https://cloudpathlib.drivendata.org/stable/caching/#automatically). Two `convert` calls are made, both using the default temporary cache location to store the SQLite file locally. When the first `convert` completes, the second `convert` seems to begin the process of reading from the same cache location which seems to persist for a short time after the first `convert` completes. Eventually, the temporary directory supplied from Cloudpathlib is removed while the second `convert` is running. I hypothesize that tests only periodically fail due to how quickly they can generally run, with Python 3.8 (where the issue often occurs) missing performance improvements available to Python 3.9 and 3.10.

Detection of the missing source data file has gone undetected due to a missing `raise` for DuckDB exceptions when extracting data. A `raise` is added for these exceptions with #82 (where linting demonstrated a need through `_get_table_columns_and_types` changes during development). A Github Actions job run which exhibits the error with a `raise` [may be found here](https://github.com/d33bs/CytoTable/actions/runs/5728135298/job/15522026550). Full logs are attached here [here](https://github.com/cytomining/CytoTable/files/12234217/1_run_tests.3.8.txt) for later reference in case they're purged.

After adding an explicit `local_cache_dir` value for these test runs of `convert` within test_convert_s3_path_sqlite I no longer have witnessed the issue occurring.

Thanks in advance for any thoughts and feedback on this PR!

# Checklist

### DIFF
--- a/tests/test_convert.py
+++ b/tests/test_convert.py
@@ -6,8 +6,6 @@ Tests for CytoTable.convert and related.
 
 import itertools
 import pathlib
-import shutil
-import tempfile
 from shutil import copy
 from typing import Any, Dict, List, Tuple, cast
 
--- a/tests/test_convert.py
+++ b/tests/test_convert.py
@@ -595,7 +595,7 @@ def test_convert_s3_path_sqlite(
         source=convert(
             source_path="s3://example/nf1/",
             dest_path=(
-                f"{get_tmpdir}/{pathlib.Path(data_dir_cellprofiler_sqlite_nf1).name}"
+                f"{get_tempdir}/{pathlib.Path(data_dir_cellprofiler_sqlite_nf1).name}"
                 ".cytotable.parquet"
             ),
             dest_datatype="parquet",
--- a/tests/test_convert.py
+++ b/tests/test_convert.py
@@ -548,6 +548,7 @@ def test_convert_s3_path_csv(
 
 
 def test_convert_s3_path_sqlite(
+    get_tempdir: str,
     data_dir_cellprofiler_sqlite_nf1: str,
     example_s3_endpoint: str,
 ):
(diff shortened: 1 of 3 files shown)


tests/test_convert.py
```diff
@@ -548,6 +546,7 @@ def test_convert_s3_path_csv(
 
 
 def test_convert_s3_path_sqlite(
+    get_tempdir: str,
```
gwaybio, line 549: `get_tempdir` is a bit confusing to me with a `str` type hint. `get_tempdir` seems like a function name. Is there a simpler name? what was wrong with `tmpdir`?

gwaybio (approved):
One comment that need not delay merging. Thanks!

tests/test_convert.py
d33bs: Thanks for this comment - this could use some improvement and the naming was mostly of convenience at the time it was created. The name stems from a fixture which generates and afterwards destroys temporary directories for tests [found here](https://github.com/cytomining/CytoTable/blob/main/tests/conftest.py#L34). When the fixture is called this way through this test a function runs to yield a new temporary directory path string and destroys that temporary directory when the test completes (post yield). The design here is to help with testing isolation and parallel performance improvements I hope we can eventually reach (where each test should not share access to a single/central temporary directory).

I naively thought this might split the definition from that of what's found under native [`tempfile`](https://docs.python.org/3/library/tempfile.html) names. I'll be more explicit here by adding an `fx_` in an attempt to make this more distinguishable to this package's work while also calling out that a function runs to generate this string through a pytest fixture.

Issue #86, opened 2023-08-14
Using `convert` to merge only two compartments causes either data loss or data mismatching


jenna-tomkinson:
# Problem

I have tried many different variations of convert to avoid this issue. I will provide one example in this issue.

To explain, I have a SQLite file acquired directly from CellProfiler but the data within the file is different from the traditional image-based analysis pipeline where there are three object compartments (nuclei, cells, and cytoplasm).

In the case of the project I am working on, I only have one object compartment which is nuclei along with whole image features. I tried using the `convert` function to merge the nuclei features and image features + metadata using the below code:

```python
# Only select from the two compartments that we have and merge them
preset_join = """
SELECT
    *
FROM
    read_parquet('per_image.parquet') as per_image
INNER JOIN read_parquet('per_nuclei.parquet') AS per_nuclei ON
    per_nuclei.Metadata_ImageNumber = per_image.Metadata_ImageNumber
"""

# merge single cells and output as parquet file
    convert(
        source_path=str(plate_folder),
        dest_path=str(output_path),
        dest_datatype=dest_datatype,
        metadata=["image"],
        compartments=["nuclei"],
        identifying_columns=["ImageNumber"],
        joins=preset_join,
        parsl_config=local_htex,
        chunk_size=10000
    )
```

The data is a 384 well plate with 4 FOVs per well, meaning there are a total of 1536 image sets. But everytime I have tried running the `convert` function with different parameter, I will get the correct number of rows of nuclei based on the SQLite database compartment "per_nuclei", but I will have missing Wells and ImageNumbers.

For example, I have a converted parquet plate loaded in a dataframe that shows 256,074 rows which matches the SQLite file:
 
![image](https://github.com/cytomining/CytoTable/assets/107513215/4c986469-c375-405a-95e1-a51e13ba1be0)

This makes sense, but for a sanity check, I went to see if I had all of the wells and image numbers since when I skimmed the SQLite file I did not see any missing wells when looking at all of the nuclei. I see that I am missing a ton of rows and their corresponding images:

![image](https://github.com/cytomining/CytoTable/assets/107513215/8a187c0b-e366-4d47-b9bc-04427351b266)
![image](https://github.com/cytomining/CytoTable/assets/107513215/d8dad460-26f9-45b0-adfb-ae9b3a207754)

[...]

Issue #85, opened 2023-08-12
Chunk_size parameter when using with preset + joins variables causes issue with `Metadata_ImageNumber`


jenna-tomkinson:
# Problem

When using the `chunk_size` parameter above 1000 with my code, it is causing the "ImageNumber" column in the per_nuclei.parquet file to start from 695 and not 1. This causes the file after concat to only have 695 to the last image number, but it has the correct number of rows for nuclei compared to the SQLite file. I had to change from LEFT JOIN to INNER so that I could get the correct number of rows based on the SQLite file. Using LEFT or RIGHT caused chaos because it was generating extra rows where the nuclei features were all NaN.

Here is the code I was using:

```python
# Only select from the two compartments that we have and merge them
preset_join = """
SELECT
    *
FROM
    read_parquet('per_image.parquet') as per_image
INNER JOIN read_parquet('per_nuclei.parquet') AS per_nuclei ON
    per_nuclei.Metadata_ImageNumber = per_image.Metadata_ImageNumber
"""

for plate_folder in sqlite_dir.iterdir():
    output_path = pathlib.Path(f"{output_dir}/{plate_folder.stem}_converted.parquet")
    # merge single cells and output as parquet file
    convert(
        source_path=str(plate_folder),
        dest_path=str(output_path),
        dest_datatype=dest_datatype,
        preset=preset,
        joins=preset_join,
        parsl_config=local_htex,
        chunk_size=10000
    )
    break
```
# Solution

If I don't use the chunk_size parameter, it looks like it is fixed. But, it goes from running for 1 minute to 8 minutes without the chunk_size increased. I am not sure what is going on at all.

Please feel free to change the title to make it more concise and I am happy to explain and demonstrate with my dataset what is happening.

jenna-tomkinson:
# Update

It looks like when using the `unique()` function, it was printing all of the unique values in the column "Metadata_ImageNumber" in the same order as the row index, so that is why I saw 695 at the start:

![image](https://github.com/cytomining/CytoTable/assets/107513215/e4a6d5eb-efdd-4293-a451-c1a96c93f495)

Once I sorted in order by the "Metadata_ImageNumber" column, I could see that the values for this did not start from 695 and leave out all numbers before:

![image](https://github.com/cytomining/CytoTable/assets/107513215/21a525d7-1016-40ab-85ef-4b35e268c794)

[...]

PR #87, opened 2023-08-16
Move to PyArrow parquet writer for data extraction

d33bs:
# Description

This PR seeks to address concerns which were raised by @jenna-tomkinson via #85 and #86. There appears to be a challenge in extracting certain data from SQLite through DuckDB and writing it directly to a file with DuckDB. This change updates CytoTable to use DuckDB for extraction (remains the same) and PyArrow's Parquet table writer for writing to file (updated). In testing this functionality with currently non-public datasets I was able to observe the correct number of unique ImageNumber keys when using only an Image and Nuclei table with SQL inner joins (as described in the issues).

# Checklist

### DIFF
--- a/cytotable/convert.py
+++ b/cytotable/convert.py
@@ -312,16 +312,18 @@ def _source_chunk_to_parquet(
     # with exception handling to read mixed-type data
     # using sqlite3 and special utility function
     try:
-        # isolate using new connection to read data with chunk size + offset
-        # and export directly to parquet via duckdb (avoiding need to return data to python)
-        _duckdb_reader().execute(
-            f"""
-            COPY (
+        # read data with chunk size + offset
+        # and export to parquet
+        parquet.write_table(
+            table=_duckdb_reader()
+            .execute(
+                f"""
                 {base_query}
                 LIMIT {chunk_size} OFFSET {offset}
-            ) TO '{result_filepath}'
-            (FORMAT PARQUET);
-            """
+                """
+            )
+            .arrow(),
+            where=result_filepath,
         )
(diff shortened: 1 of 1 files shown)


falquaddoomi (approved):
I'm unfortunately not familiar enough with either DuckDB or SQLite to evaluate the underlying difference, but the changes here look modest and reasonable.

gwaybio (approved):
Two minor questions, which may not require any changes and should not delay merging.

1. Do you need to update the documentation above the `try` statement (lines 311-313) to reflect the parquet writer change?
2. I imagine that the `except` clause would catch when the `_duckdb_reader()` fails, so probably no need to change this clause alongside the `try` logic change?

d33bs:
Thanks @gwaybio and @falquaddoomi for the reviews! Addressing questions below:

> 1. Do you need to update the documentation above the `try` statement (lines 311-313) to reflect the parquet writer change?

Thank you, I've updated this to be a bit more explicit, also moving around some of the mentions to appropriate sections of the code.

> 2. I imagine that the `except` clause would catch when the `_duckdb_reader()` fails, so probably no need to change this clause alongside the `try` logic change?

Thank you for checking! Any exceptions would still emerge here when they're encountered. The "mixed-type" DuckDB exception traces to the nested select within the `COPY` functionality and this will persist in the new format as well (the `SELECT` query doesn't change). This functionality is tested within [`test_cell_health_cellprofiler_to_cytominer_database_legacy`](https://github.com/cytomining/CytoTable/blob/main/tests/test_convert.py#L1087), which utilizes a dataset that includes a SQLite string value within a float column.

Issue #90, opened 2023-08-21
Cytotable convert fails to reload config file outside `for` loop


jenna-tomkinson:
# Problem

When creating a parsl config variable outside of a for loop running `convert` on multiple SQLite files, it is not able to reload the config when running on the next file producing this error:

```
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
File ~/mambaforge/envs/durbin_lab_env/lib/python3.8/site-packages/cytotable/convert.py:1338, in convert(source_path, dest_path, dest_datatype, source_datatype, metadata, compartments, identifying_columns, concat, join, joins, chunk_columns, chunk_size, infer_common_schema, drop_null, data_type_cast_map, preset, parsl_config, **kwargs)
   1336     else:
   1337         # else we attempt to load the given parsl configuration
-> 1338         parsl.load(parsl_config)
   1339 except RuntimeError as runtime_exc:
   1340     # catch cases where parsl has already been loaded and defer to
   1341     # previously loaded configuration with a warning

File ~/mambaforge/envs/durbin_lab_env/lib/python3.8/site-packages/typeguard/__init__.py:1033, in typechecked..wrapper(*args, **kwargs)
   1032 check_argument_types(memo)
-> 1033 retval = func(*args, **kwargs)
   1034 try:

File ~/mambaforge/envs/durbin_lab_env/lib/python3.8/site-packages/parsl/dataflow/dflow.py:1423, in DataFlowKernelLoader.load(cls, config)
   1422 if cls._dfk is not None:
-> 1423     raise RuntimeError('Config has already been loaded')
   1425 if config is None:

RuntimeError: Config has already been loaded

During handling of the above exception, another exception occurred:

KeyError                                  Traceback (most recent call last)
...
    339                                start_method=self.start_method)
    340 self.launch_cmd = l_cmd
    341 logger.debug("Launch command: {}".format(self.launch_cmd))

KeyError: 'block_id'
```

I am using this code as an example for getting this error:

```python
# parsl workflow executer to run as multi-processing instead of multi-threaded (found in cytotable)
local_htex = Config(
    executors=[
        HighThroughputExecutor(
            label="htex_Local",
            worker_debug=True,
            cores_per_worker=1,
            provider=LocalProvider(
                channel=LocalChannel(),
                init_blocks=1,
                max_blocks=1,
            ),
        )
    ],
    strategy=None,
    )

for plate_folder in sqlite_dir.iterdir():
[...]

PR #93, opened 2023-08-24
Reuse existing Parsl configurations

d33bs:
# Description

This PR updates CytoTable to utilize existing Parsl configurations when using `convert` multiple times in the same process. This change comes about through recommendations from parsl/parsl#2871 to reuse Parsl configurations and defer to external configuration where possible.

References #90 

# Checklist

### DIFF
--- a/cytotable/convert.py
+++ b/cytotable/convert.py
@@ -1314,8 +1314,8 @@ def convert(  # pylint: disable=too-many-arguments,too-many-locals
             an optional group of presets to use based on common configurations
         parsl_config: Optional[parsl.Config] (Default value = None)
             Optional Parsl configuration to use for running CytoTable operations.
-            Note: when using CytoTable multiple times in the same process the
-            first configuration loaded will be reused during subsequent runs.
+            Note: when using CytoTable multiple times in the same process,
+            CytoTable will use the first provided configuration for all runs.
 
     Returns:
         Union[Dict[str, List[Dict[str, Any]]], str]
@@ -1358,7 +1358,7 @@ def convert(  # pylint: disable=too-many-arguments,too-many-locals
             )
     """
 
-    # attempt to load parsl configuration if we don't already have one
+    # attempt to load parsl configuration if we didn't already load one
     if not _parsl_loaded():
         # if we don't have a parsl configuration provided, load the default
         if parsl_config is None:


cytotable/convert.py
```diff
@@ -1314,6 +1314,8 @@ def convert(  # pylint: disable=too-many-arguments,too-many-locals
             an optional group of presets to use based on common configurations
         parsl_config: Optional[parsl.Config] (Default value = None)
             Optional Parsl configuration to use for running CytoTable operations.
+            Note: when using CytoTable multiple times in the same process the
+            first configuration loaded will be reused during subsequent runs.
```
gwaybio, line 1318:
```suggestion
            Note: when using CytoTable multiple times in the same process,
            CytoTable will use the first provided configuration for all runs.
```

I made some edits to increase clarity, but it's possible my suggestion doesn't quite hit the mark, so feel free to reject/edit.

cytotable/convert.py
```diff
@@ -1356,31 +1358,17 @@ def convert(  # pylint: disable=too-many-arguments,too-many-locals
             )
     """
 
-    # attempt to load parsl configuration
-    try:
+    # attempt to load parsl configuration if we don't already have one
```
gwaybio, line 1361:
```suggestion
    # attempt to load parsl configuration if we didn't already load one
```

cytotable/utils.py
```diff
@@ -9,7 +9,7 @@
 from typing import Any, Dict, Union, cast
 
 import duckdb
-import pyarrow as pa
+import parsl
```
gwaybio, line 12: Just confirming that you no longer wish to import `pyarrow`

cytotable/utils.py
```diff
@@ -99,6 +99,24 @@ def Parsl_AppBase_init_for_docs(self, func, *args, **kwargs):
...
+    try:
+        # try to reference Parsl dataflowkernel
+        parsl.dfk()
+    except RuntimeError as rte:
+        # if we detect a runtime error that states we need to load config
+        # return false to indicate parsl config has not yet been loaded.
+        if str(rte) == "Must first load config":
+            return False
+
+    # otherwise we indicate parsl config has already been loaded
```
gwaybio, line 116: Just confirming that this is designed to only catch one kind of runtime error and all others will be handled differently?

gwaybio (approved):
Great to track this down! I have only minor comments/suggestions that are optional and need not delay merging.

cytotable/utils.py
benclifford: I think this is catching this error:

> parsl/dataflow/dflow.py:            raise RuntimeError('Must first load config')

which I think parsl should probably make its own error type to support this kind of detection without looking at the error text.

I made an issue https://github.com/Parsl/parsl/issues/2873

cytotable/convert.py
d33bs: Thanks - I think this is a lot more clear! I'll add this suggestion as a commit.

cytotable/utils.py
d33bs: Thanks for the double check here. I noticed this dependency had been specified deeper in this module as I made the edits in this PR and decided to clean it up. Specifically, `_sqlite_mixed_type_query_to_parquet` has this dependency, but the import is now specified at a function-level in abidance with the style of Parsl apps (where the function is referenced).

cytotable/utils.py
d33bs: Thanks @gwaybio ! Yes, this is only catching Parsl configuration load exceptions as an indication of whether configurations have already been loaded. I'm adding a `raise` here for other `RuntimeExceptions`.

Thanks as well @benclifford ! I've added an issue to help make the updates here once there is a custom Exception: https://github.com/cytomining/CytoTable/issues/94 .

PR #109, opened 2023-09-19
Use Parsl ConfigurationError for detecting existing Parsl configurations

### LINKED ISSUE #94: Catch Parsl configuration load Exception class in `_parsl_loaded`
This issue pertains to a possible future state where Parsl includes an Exception which may be used to detect whether no configurations have been loaded. Movement on this issue can take place after https://github.com/Parsl/parsl/issues/2873 has been addressed. Discussion surrounding this topic came up via https://github.com/cytomining/CytoTable/pull/93#discussion_r1304778648 .

d33bs:
# Description

This PR updates `cytotable.utils._parsl_loaded` to use Parsl's ConfigurationError to help detect previously loaded configurations as part of a new Parsl release. More details can be found within #94 and https://github.com/cytomining/CytoTable/issues/90#issuecomment-1726359298 .

Many thanks go to @sbamin for highlighting the issue here: https://github.com/cytomining/CytoTable/issues/90#issuecomment-1726331209

References #94 

# Checklist

### DIFF
--- a/cytotable/utils.py
+++ b/cytotable/utils.py
@@ -111,7 +111,7 @@ def _parsl_loaded() -> bool:
     except ConfigurationError as pce:
         # if we detect a Parsl ConfigurationError that states we need to load config
         # return false to indicate parsl config has not yet been loaded.
-        if str(pce) == "Must first load config":
+        if pce.args[0] == "Must first load config":
             return False
 
         # otherwise we raise other ConfigurationError's
--- a/cytotable/utils.py
+++ b/cytotable/utils.py
@@ -14,6 +14,7 @@ from cloudpathlib import AnyPath, CloudPath
 from cloudpathlib.exceptions import InvalidPrefixError
 from parsl.app.app import AppBase
 from parsl.config import Config
+from parsl.errors import ConfigurationError
 from parsl.executors import HighThroughputExecutor
 
 logger = logging.getLogger(__name__)
@@ -107,13 +108,13 @@ def _parsl_loaded() -> bool:
     try:
         # try to reference Parsl dataflowkernel
         parsl.dfk()
-    except RuntimeError as rte:
-        # if we detect a runtime error that states we need to load config
+    except ConfigurationError as pce:
+        # if we detect a Parsl ConfigurationError that states we need to load config
         # return false to indicate parsl config has not yet been loaded.
-        if str(rte) == "Must first load config":
+        if str(pce) == "Must first load config":
             return False
 
-        # otherwise we raise other RuntimeError's
+        # otherwise we raise other ConfigurationError's
         else:
             raise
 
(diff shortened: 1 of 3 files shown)


cytotable/utils.py
```diff
@@ -107,13 +108,13 @@ def _parsl_loaded() -> bool:
     try:
         # try to reference Parsl dataflowkernel
         parsl.dfk()
-    except RuntimeError as rte:
-        # if we detect a runtime error that states we need to load config
+    except ConfigurationError as pce:
+        # if we detect a Parsl ConfigurationError that states we need to load config
         # return false to indicate parsl config has not yet been loaded.
-        if str(rte) == "Must first load config":
+        if str(pce) == "Must first load config":
```
gwaybio, line 114: This seems a bit brittle to me. Maybe ok for now, but does `ConfigurationError` have a text attribute to check?

gwaybio (approved):
One minor comment that need not delay merge

cytotable/utils.py
d33bs: Thank you for this, I'll move to use `pce.args[0]` which represents the string message sent to the Exception. I explored the use of [`traceback`](https://docs.python.org/3/library/traceback.html) for more isolated, declarative treatment here but I couldn't justify a switch there as less complex.

cytotable/utils.py
```diff
@@ -107,13 +108,13 @@ def _parsl_loaded() -> bool:
     try:
         # try to reference Parsl dataflowkernel
         parsl.dfk()
-    except RuntimeError as rte:
-        # if we detect a runtime error that states we need to load config
+    except ConfigurationError as pce:
+        # if we detect a Parsl ConfigurationError that states we need to load config
         # return false to indicate parsl config has not yet been loaded.
-        if str(rte) == "Must first load config":
+        if pce.args[0] == "Must first load config":
```
falquaddoomi, line 114: I looked at the Parsl source and, sure enough, there are a number of places where `ConfigurationError` is raised that are differentiated just by the exception's message text. While I also don't love checking the message text, since the Parsl maintainers could decide to tweak the text at any time and this would break, I don't see any way around it currently. Perhaps it's something to raise with the Parsl devs?

falquaddoomi (approved):
I assume this PR is mostly just catching up to recent Parsl changes. It's a little disappointing that Parsl doesn't have more specific exception types, but IMHO for now this PR is good as-is.

cytotable/utils.py
d33bs: Great points, thanks! I reached out to Parsl devs via Parsl/parsl#2873 for a [suggested approach](https://github.com/Parsl/parsl/issues/2873#issuecomment-1726540856). I also updated the PR description to reference and not close #94 (as this would suggest we need a possible further update).

d33bs:
Thank you @gwaybio and @falquaddoomi for your reviews! Thank you again @sbamin for raising this in your comments! Merging this in now.

PR #112, opened 2023-09-26
Use Parsl NoDataFlowKernelError for detecting existing Parsl configurations

d33bs:
# Description

This is a follow up to #109 to use a different Exception type for checking already loaded Parsl configurations after a recent change in Parsl v2023.09.25 . The changed exception type adds greater specificity here and I felt it would still be wise to make sure the str matches as intended.

See: https://github.com/Parsl/parsl/pull/2890 and https://github.com/Parsl/parsl/issues/2873 for more detail.

# Checklist


gwaybio (approved):
Sorry for my delay!

falquaddoomi (approved):
Same, sorry for the delay. This looks good, nice to see that the devs on the parsl end are receptive to suggestions!

One question: is it still necessary to check the exception text, or is this exception specific enough to not need that anymore? IMHO this question shouldn't block merging; I'm mostly curious.

d33bs:
Thank you @gwaybio and @falquaddoomi for the reviews! @falquaddoomi , addressing your question: I mostly kept the string as a safety in case `NoDataFlowKernelError` is used in other locations within Parsl moving forward. It could be used in multiple spots in the future with different messages, so I figured we could keep the string-based check as part of this for now to be "more" certain of the circumstances when/if things change.

Just the same, to my knowledge, this is the only case where the exception may be emitted currently from Parsl, so maybe it's best to remove it to keep things flexible to the future. There might be a danger of the exception being used in other areas, but maybe this danger is less than possibility of the string message changing. I'm unsure what those probabilities might be, but defer towards flexibility, and will make a change accordingly.

PR #158, opened 2024-02-10
Changing python's upper limit version to python<3.12

axiomcura:
# Description

Thank you for your contribution to CytoTable!
Please _succinctly_ summarize your proposed change.
What motivated you to make this change?

This PR addresses the current compatibility issues between `duckdb` and Python 3.12. As `duckdb` does not yet support Python 3.12, it is best to downgrade the upper limit of the Python version in the environment poetry.lock file. 

Below is the current issue 

Issue mentioned: https://github.com/duckdb/duckdb/issues/9301#issuecomment-1756731027
Current PR (however closed) : https://github.com/duckdb/duckdb/pull/9222 

Changes:
* Downgrade the upper limit Python version to Python <3.12 to maintain compatibility with `duckdb`.

Additionally, I have written an issue explaining why Python >3.12 does not work with DuckDB, which directly impacts the installation of CytoTable. #157 

Please also link to any relevant issues that your code is associated with.

## What is the nature of your change?

- [X] Bug fix (fixes an issue).
- [ ] Enhancement (adds functionality).
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected).
- [ ] This change requires a documentation update.

# Checklist

Please ensure that all boxes are checked before indicating that a pull request is ready for review.

- [X] I have read the [CONTRIBUTING.md](CONTRIBUTING.md) guidelines.
- [X] My code follows the style guidelines of this project.
- [X] I have performed a self-review of my own code.
- [ ] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have made corresponding changes to the documentation.
- [X] My changes generate no new warnings.
- [ ] New and existing unit tests pass locally with my changes.
- [ ] I have added tests that prove my fix is effective or that my feature works.
- [ ] I have deleted all non-relevant text in this pull request template.

Issue #143, opened 2024-01-11
`parsl.jobs.errors.TooManyJobFailuresError` sporadically occurs on MacOS Images for GitHub Actions jobs during `pytest` operations


d33bs:
This issue focuses on a sporadic issue witnessed on the MacOS image while running `pytest` tests for CytoTable from GitHub Actions with a repeated error message:

```text
parsl.jobs.errors.TooManyJobFailuresError: Error 1:
	EXIT CODE: 5
	STDOUT: Found cores : 4
Launching worker: 1
Failed to find a viable address to connect to interchange. Exiting
```

On rerunning the job after changes where it was witnessed, the error doesn't occur again. I'm unsure if this is related to the GitHub Actions platform, Parsl, how Parsl is used by CytoTable, or some combination of all three. 

Reference job where this occurs: https://github.com/cytomining/CytoTable/actions/runs/7480639908/job/20360513109

d33bs:
Adding another example from @kenibrewer in #164
> @d33bs Any insight into why [this testing run](https://github.com/cytomining/CytoTable/actions/runs/8290119489/job/22687673141#step:8:13056) failed, but [this re-run ](https://github.com/cytomining/CytoTable/actions/runs/8290119489/job/22703583556?pr=164)succeeded?

d33bs:
Also occurred temporarily in #196 via https://github.com/cytomining/CytoTable/actions/runs/8806919576/job/24172783425 . Re-running seemed to succeed. Earlier versions of the `macos-latest` image runner tag I believe referenced `macos-12` whereas #196 is intentionally using `macos-13` (same issue appears to cross the major versions).

PR #164, opened 2024-03-14
Add and apply pre-commit poetry checks

### LINKED ISSUE #159: Add Poetry pre-commit check for lockfile updates
Poetry provides a [pre-commit check for lockfiles](https://python-poetry.org/docs/master/pre-commit-hooks/#poetry-lock) which can automatically account for lockfile updates.

Related discussion:

> @d33bs we experimented with some of the poetry precommit checks right? I
can't remember which ones we used and why we removed them. I've been using
the poetry lock hook in other repos more recently and it works well and
would catch this pretty common miss.

> Thanks @kenibrewer ! I think it's a good idea to include Poetry pre-commit checks for the lockfile! I have a vague recollection there may have been slight inconsistency between Poetry lockfiles from OS to OS (for example, MacOS vs Ubuntu Linux), incurring an "endless loop" of pre-commit check fix-failures. Either way, it'd be great to have this functionality if we can get it working! I'll add an issue to this effect.

_Originally posted by @kenibrewer in https://github.com/cytomining/CytoTable/issues/158#issuecomment-1942222418_

d33bs:
# Description

This adds a pre-commit check which automatically checks the Poetry-based `pyproject.toml` file for formatting and the `poetry.lock` file for related updates.

I noticed the check for the lockfile was failing due to SSH errors related to the git-based hash for Pycytominer. As a result, I opted to update pycytominer to use the latest PyPI release of `1.1.0`. This will likely speed up installations more widely.

Closes #159 
Relates to #158 (CC @axiomcura)

# Checklist

- [ ] I have commented my code, particularly in hard-to-understand areas.


d33bs:
This change will likely depend on #166 .

gwaybio (approved):

.pre-commit-config.yaml
```diff
@@ -11,6 +11,11 @@ repos:
       - id: check-yaml
       - id: check-added-large-files
       - id: check-toml
+  - repo: https://github.com/python-poetry/poetry
+    rev: 1.8.0
+    hooks:
+      - id: poetry-check
+      - id: poetry-lock
```
kenibrewer, line 18: I'm in favor of including `poetry-check`, but I'm uncertain whether or not we want to be including `poetry-lock` as well. Because `poetry-lock` checks to see whether dependencies are the "most up to date", it could run differently on a user's machine vs later when the same commit gets pushed to github. I think instead we should rely on dependabot for package updates.

kenibrewer (changes requested):
Marking this as request changes to highlight an important item of discussion below regarding `poetry-lock`.

.pre-commit-config.yaml
kenibrewer: Alternately, we could add the `--no-update` argument to the `poetry-lock` check

kenibrewer:
@d33bs Any insight into why [this testing run](https://github.com/cytomining/CytoTable/actions/runs/8290119489/job/22687673141#step:8:13056) failed, but [ this re-run ](https://github.com/cytomining/CytoTable/actions/runs/8290119489/job/22703583556?pr=164) succeeded?

d33bs:
Thanks @gwaybio and @kenibrewer ! I've moved to using `poetry-check` as suggested to avoid potential issues.

> @d33bs Any insight into why [this testing run](https://github.com/cytomining/CytoTable/actions/runs/8290119489/job/22687673141#step:8:13056) failed, but [this re-run ](https://github.com/cytomining/CytoTable/actions/runs/8290119489/job/22703583556?pr=164)succeeded?

I believe this is roughly the same issue as reported in #143 . I don't know for sure, but my hunch is that it has something to do with multiprocessing from Parsl used through the non-threaded tests in CytoTable alongside potential nondeterministic constraints by which MacOS environments are provisioned from GitHub for Actions workflows. I don't have any proof of this, but the fact that it only happens sporadically on MacOS virtual environments has me imagine as though they are (perhaps rarely, not guaranteed) shared resources across multiple workflows.

kenibrewer (approved):
LGTM

Thanks for the insight about the test failure!

d33bs:
Thanks as well @kenibrewer ! Merging this in.

PR #167, opened 2024-03-14
Add weekly testing for preventative awareness of external changes

d33bs:
# Description

This PR adds a weekly check occurring at 7 AM UTC on Wednesdays to help detect external changes that could effect CytoTable (such as dependency updates). This change comes as a result of the issues in #163, which were discovered by @shntnu in the process of attempting to use CytoTable. Generally, I expect this to happen again with other dependencies over time (I believe we've also encountered similar issues with Parsl, for example), so this is an attempt to add a simple solution towards prevention. After this change completes, my hope would be that we might have early awareness of these types of issues in the future such that users don't have as much burden to troubleshoot.

References #163

## What is the nature of your change?

- [ ] Bug fix (fixes an issue).
- [x] Enhancement (adds functionality).
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected).
- [ ] This change requires a documentation update.

# Checklist

Please ensure that all boxes are checked before indicating that a pull request is ready for review.

- [x] I have read the [CONTRIBUTING.md](CONTRIBUTING.md) guidelines.
- [x] My code follows the style guidelines of this project.
- [x] I have performed a self-review of my own code.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have made corresponding changes to the documentation.
- [x] My changes generate no new warnings.
- [x] New and existing unit tests pass locally with my changes.
- [ ] I have added tests that prove my fix is effective or that my feature works.
- [x] I have deleted all non-relevant text in this pull request template.

Issue #163, opened 2024-03-12
ArrowInvalid because of differing schemas upon join


shntnu:
I get this error [cytotable_error.txt](https://github.com/cytomining/CytoTable/files/14565739/cytotable_error.txt) when running the code below.

There's something funky happening; from the log:

```
ArrowInvalid: Schema at index 0 was different:
Metadata_TableNumber: string
Metadata_ImageNumber: int64
Image_Metadata_Well: string
Image_Metadata_Plate: string
Metadata_TableNumber: string <-----
Metadata_ImageNumber: int64 <-----
Metadata_ObjectNumber: int64
...
vs
Metadata_TableNumber: string
Metadata_ImageNumber: int64
Image_Metadata_Well: string
Image_Metadata_Plate: string
Metadata_TableNumber_1: string <-----
Metadata_ImageNumber_1: int64 <-----
Metadata_ObjectNumber: int64
...
```

```py
# Author: Zitong (Sam) Chen, Broad Institute, 2023
#
# Download sample SQLite file:
# wget https://raw.githubusercontent.com/d33bs/pycytominer/43cf984067700aa52f0b6752e3490d9e12d60170/tests/test_data/cytominer_database_example_data/test_SQ00014613.sqlite -O test_SQ00014613.sqlite
# Examine column names:
# sqlite3 test_SQ00014613.sqlite .schema|grep -v AreaShape|grep -vE "Texture|RadialDistribution|Granularity|Location|Neighbors|Correlation|Intensity|ObjectSkeleton|ExecutionTime|ImageQuality|ModuleError|PathName|MD5Digest|Scaling|Width|FileName|URL|Threshold|Height|Count|CREATE INDEX|IncludingEdges"
# Output:
# CREATE TABLE IF NOT EXISTS "Image" (
#         "TableNumber" TEXT,
#         "Image_Group_Index" BIGINT NOT NULL,
#         "Image_Group_Number" BIGINT NOT NULL,
#         "ImageNumber" BIGINT NOT NULL,
#         "Image_Metadata_AbsPositionZ" FLOAT NOT NULL,
#         "Image_Metadata_AbsTime" DATETIME,
#         "Image_Metadata_BinningX" BIGINT NOT NULL,
#         "Image_Metadata_BinningY" BIGINT NOT NULL,
#         "Image_Metadata_ChannelID" BIGINT NOT NULL,
#         "Image_Metadata_ChannelName" TEXT,
[...]

PR #170, opened 2024-03-18
Remove poetry lockfile to better project testing with scheduled trigger

d33bs:
# Description

This PR is an addendum to #167 to assist with scheduled projected testing. The intent here would be to remove the lockfile in case there are dependencies which users may see upstream that we don't yet see in CytoTable testing due to the lockfile. For example, in the case of #163 testing would not have failed for scheduled tests due to the lockfile avoiding the newer versions of `duckdb`.

Thanks for any thoughts or feedback you may have on this!

# Checklist

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -27,6 +27,12 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python_version }}
+      # remove poetry.lock file for scheduled tests
+      # to help simulate possible upstream issues
+      - name: Remove poetry.lock for scheduled tests
+        if: github.event_name == 'schedule'
+        run: |
+          rm poetry.lock
       - name: Setup for poetry
         uses: ./.github/actions/setup-poetry
       - name: Install environment


gwaybio (approved):

kenibrewer:
Hmmm.... I'm not sure removing the the lock file is the appropriate fix here. I'll dig into this a bit later with my rationale.

d33bs:
Thanks @gwaybio and @kenibrewer ! More background, in case it helps, what I've been seeing lately is that we incur "surprise" dependency related challenges due to the `poetry.lock` file and the differences in time alignment between PR's vs external release timelines. This is intended to be a proactive and observable way to detect these possibly before users experience them. Ideally, if we can catch them and fix them quickly enough, users would never see these challenges in their own work.

kenibrewer:
Here's what I'm confused about. According to the [description of the poetry lock file](https://python-poetry.org/docs/basic-usage/#installing-with-poetrylock), the lock file is intended to prevent exactly this type of scenario that occurred with #163. It provides a specific list of allowed versions for each dependency. 

It might be helpful understand a bit more about how exactly the incompatible version of duckdb was installed. Did the poetry.lock get updated to allow duckdb 0.10.0 even though we didn't test against it? Is the information from poetry.lock file not kept when building the package for pip and/or conda?

d33bs:
Hi @kenibrewer , thanks for your reply on this. The user who reported #163 experienced the issue because of a loose constraint on `duckdb` from within the `pyrproject.toml` file which they received through a non-`poetry.lock`'ed installation. The `poetry.lock` file did its job within the context of the source code for CytoTable, providing a reproducible testing experience. However, the user still ran into the issue and opted to report the challenge to us for awareness and a fix. The issue occurred because external dependencies are updated outside of existing triggers for CytoTable.

This change leans towards CytoTable developers having a proactive awareness of issues like the one experienced in #163 . It's an experiment towards improved user experience through increased visibility (and an implied developer responsibility as a result).

kenibrewer (approved):
@d33bs Thanks for the additional clarification. I didn't realize that only the version constraints from the pyproject.toml are used in some downstream installation methods. LGTM

d33bs:
Thanks @kenibrewer !

PR #174, opened 2024-03-22
Enable Python 3.12 compatibility through dependency-based updates

### LINKED ISSUE #157: NOTE: Compatibility Issue with Python 3.12 and CytoTable Due to Incompatibilities with DuckDB
## current issues with python 3.12 
There is a compatibility issue with Python 3.12 which causes CytoTable to fail when installing. This primarily stemming from incompatibilities with DuckDB.

### Context
When trying to install CytoTable using Poetry, an error pops up saying DuckDB doesn't follow PEP 517 standards. This  prevents `CytoTable` being installed through pip, poetry and other software packages that use CyoTable as a dependency. 

Digging into the DuckDB github issue forums, there's an  [issue](https://github.com/duckdb/duckdb/issues/9301#issuecomment-1756731027) that points to incompatibilities with Python 3.12. Users are seeing similar errors and behavior when attempting to install DuckDB with Python 3.12. The forum discussion shows the DuckDB team knows about the issue and is trying to fix it by making some changes. 
- [duckdb-pr](https://github.com/duckdb/duckdb/pull/9222)

Currently, the issue remains unresolved. As a workaround, we decided to adjusting the upper limit version of Python from `<3.13` to ` <3.12`. This small change allows duckdb to be installed properly within the CytoTable's environment.

d33bs:
# Description

This PR enables Python 3.12 compatibility with CytoTable through various dependency-based updates. The original work focused on fixes for #157 and #163 via `duckdb==0.10.1`. I found that Python 3.12 also required changes to `numpy`, `scipy`, and `cloudpathlib` in the form of updates or Python to dependency version constraints (without these changes installation or testing through Python 3.12 fails).

In addition, I found that `duckdb==0.10.1` entails lowered determinism for joined data result value sorting. We previously have relied on PyArrow to consistently sort all columns and their values for testing - this appears to also be potentially in need of adjustment. As a result, I've added SQL [`ORDER BY ALL`](https://duckdb.org/docs/sql/query_syntax/orderby.html#order-by-all) (which orders all values in all columns from left to right) to one test which consistently failed with comparisons. As a to-do, we should explore why PyArrow is unable to perform the same level of data organization for testing and possibly raise an issue with that project if the results aren't in alignment with the design. Alternatively, moving to `ORDER BY ALL` or some parallel may be necessary to address the existing PyArrow-based data sorting for testing (otherwise we may sporadically see issues over time).

This touches also on #85, where we have questions regarding if / how / when to sort CytoTable results. Currently, CytoTable output column ordering is sorted and columns values are not sorted. It might be beneficial to add `ORDER BY ALL` as a step within CytoTable if we determine this is a feature folks would like.

A new test warning is emitted which I believe entails changes in Python 3.12 and Parsl: `DeprecationWarning: This process (pid=######) is multi-threaded, use of fork() may lead to deadlocks in the child.`. This appears to be tracked via parsl/parsl#2343 .

Closes #157

# Checklist

- [ ] My changes generate no new warnings.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -3011,41 +3011,41 @@ tests = ["black (>=23.3.0)", "matplotlib (>=3.1.3)", "mypy (>=1.3)", "numpydoc (
 
 [[package]]
 name = "scipy"
-version = "1.9.3"
+version = "1.10.0"
 description = "Fundamental algorithms for scientific computing in Python"
 optional = false
-python-versions = ">=3.8"
+python-versions = "<3.12,>=3.8"
 files = [
-    {file = "scipy-1.9.3-cp310-cp310-macosx_10_9_x86_64.whl", hash = "sha256:1884b66a54887e21addf9c16fb588720a8309a57b2e258ae1c7986d4444d3bc0"},
-    {file = "scipy-1.9.3-cp310-cp310-macosx_12_0_arm64.whl", hash = "sha256:83b89e9586c62e787f5012e8475fbb12185bafb996a03257e9675cd73d3736dd"},
-    {file = "scipy-1.9.3-cp310-cp310-manylinux_2_17_aarch64.manylinux2014_aarch64.whl", hash = "sha256:1a72d885fa44247f92743fc20732ae55564ff2a519e8302fb7e18717c5355a8b"},
-    {file = "scipy-1.9.3-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl", hash = "sha256:d01e1dd7b15bd2449c8bfc6b7cc67d630700ed655654f0dfcf121600bad205c9"},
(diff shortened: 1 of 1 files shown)


kenibrewer (approved):
Thanks for the detailed explanation. Do you know if there is a performance impact from the `ORDER BY ALL`? 

Otherwise, looks good to me.

d33bs:
Thank you @kenibrewer ! 

> Do you know if there is a performance impact from the ORDER BY ALL?

I don't have enough data on this front to be certain. My limited knowledge is that DuckDB internally leverages similar but different methodology to PyArrow when it comes to data handling. Anecdotally, I've witnessed that sometimes one outperforms the other, but that this doesn't seem to be consistent from dataset to dataset (or version to version).

I compared some of the tests from GH Actions and saw that, for example, Python 3.11 on Ubuntu testing was completed in a few seconds longer from `main` vs this PR. This could be shared resource variance from GitHub however.

PR #182, opened 2024-04-02
Add order to limit and offset queries for deterministic results

### LINKED ISSUE #175: Explore value sorting determinism (and possible changes)
From #174:

> I found that `duckdb==0.10.1` entails lowered determinism for joined data result value sorting. We previously have relied on PyArrow to consistently sort all columns and their values for testing - this appears to also be potentially in need of adjustment. As a result, I've added SQL [`ORDER BY ALL`](https://duckdb.org/docs/sql/query_syntax/orderby.html#order-by-all) (which orders all values in all columns from left to right) to one test which consistently failed with comparisons. As a to-do, we should explore why PyArrow is unable to perform the same level of data organization for testing and possibly raise an issue with that project if the results aren't in alignment with the design. Alternatively, moving to `ORDER BY ALL` or some parallel may be necessary to address the existing PyArrow-based data sorting for testing (otherwise we may sporadically see issues over time).

Because of the importance of this issue, adding that we need example cases where the fix has been validated with larger than testing datasets.

d33bs:
# Description

This PR addresses #175 by adding `ORDER BY ALL` to SQL queries which are used to help parallelize data work through `LIMIT` and `OFFSET`. This comes as the result of various testing surrounding PyArrow, DuckDB, and Parsl and reading + experimentation on what happens during DuckDB queries which leverage `LIMIT`, `OFFSET`, and lists of Parquet files treated as a single dataset. DuckDB provides documentation supporting this change for determinism: `"Note that while LIMIT can be used without an ORDER BY clause, the results might not be deterministic without the ORDER BY clause."`. ([link](https://duckdb.org/docs/sql/query_syntax/limit)). The same stands for SQLite-based queries ([link](https://www.sqlite.org/lang_select.html#orderby)).

These additions required a small changes due to the new sorting order of results.

Closes #175 
May influence work towards #85

# Checklist

### DIFF
--- a/cytotable/convert.py
+++ b/cytotable/convert.py
@@ -348,6 +348,8 @@ def _source_chunk_to_parquet(
                 table=ddb_reader.execute(
                     f"""
                     {base_query}
+                    /* order by all columns for deterministic output */
+                    ORDER BY ALL
                     LIMIT {chunk_size} OFFSET {offset}
                     """
                 ).arrow(),
@@ -750,6 +752,7 @@ def _join_source_chunk(
         result = ddb_reader.execute(
             f"""
                 {joins}
+                {"ORDER BY ALL" if "ORDER BY" not in joins.upper() else ""}
                 LIMIT {chunk_size} OFFSET {offset}
                 """
         ).arrow()
--- a/cytotable/utils.py
+++ b/cytotable/utils.py
@@ -257,7 +257,12 @@ def _sqlite_mixed_type_query_to_parquet(
 
         # perform the select using the cases built above and using chunksize + offset
         cursor.execute(
-            f'SELECT {", ".join(query_parts)} FROM {table_name} LIMIT {chunk_size} OFFSET {offset};'
+            f"""
+            SELECT {', '.join(query_parts)}
+            FROM {table_name}
+            ORDER BY {', '.join([col['column_name'] for col in column_info])}
+            LIMIT {chunk_size} OFFSET {offset};
+            """
         )
         # collect the results and include the column name with values
         results = [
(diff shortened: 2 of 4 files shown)


falquaddoomi (approved):
This seems like a reasonable change to me in terms of ensuring correctness, since as you observed `LIMIT..OFFSET` queries require a fixed ordering. From what I've read, `ORDER BY ALL` in DuckDB orders by all columns from first to last, which is the safest option since you'll always have a canonical ordering, but it could impact performance if the columns weren't indexed. I don't know much about DuckDB indexing, but from a cursory reading it seems the engine creates indices on all columns by default, so this may not be an issue.

In any case, I'd do some light checking on if or how much it would impact performance, but I don't see any reason not to approve this PR.

gwaybio (approved):
Probably best served by a different PR, but I'm wondering if it is worth adding a test that ensures, in a large enough dataset, that there are no duplicate rows and all expected rows are present.

d33bs:
Thank you @gwaybio and @falquaddoomi for your reviews! I've added #193 for future-facing large dataset testing focus.

> In any case, I'd do some light checking on if or how much it would impact performance, but I don't see any reason not to approve this PR.

I'm uncertain of the impacts this will have towards performance but found [this article](https://duckdb.org/2021/08/27/external-sorting.html) covering a bit how sorting operates within DuckDB (I'm unsure if this extends into the Parquet-focused readers, as these may operate differently based on how Parquet is stored). I plan to follow up within #175 with additional external large data tests, including source, to monitor the impact of how sorting might impact things.

Merging this in to help address next steps.

PR #183, opened 2024-04-04
Use compartment tables as basis of preset joins

### LINKED ISSUE #86: Using `convert` to merge only two compartments causes either data loss or data mismatching

d33bs:
# Description

This PR changes the way SQL-based joins in presets produce data to avoid scenarios where there is image table data but no associated compartment table data. Many thanks to @jenna-tomkinson who helped provide details and troubleshooting via #86. To help make sure we don't run into this challenge again, I've added a test using data related to #86.

This change is based off of and should be merged after #182 .

Closes #86

# Checklist

### DIFF
--- a/tests/test_convert_threaded.py
+++ b/tests/test_convert_threaded.py
@@ -239,6 +239,7 @@ def test_get_source_filepaths(
     # test that the single dir structure includes 4 unique keys
     assert len(set(single_dir_result.keys())) == 4
 
+
 def test_avoid_na_row_output(
     load_parsl_threaded: None, fx_tempdir: str, data_dir_cellprofiler: str
 ):
--- a/tests/test_convert_threaded.py
+++ b/tests/test_convert_threaded.py
@@ -10,6 +10,7 @@ from typing import Any, Dict, List, cast
 
 import parsl
 import pyarrow as pa
+import pyarrow.compute as pc
 import pytest
 from parsl.config import Config
 from parsl.executors import ThreadPoolExecutor
@@ -237,3 +238,33 @@ def test_get_source_filepaths(
     ).result()
     # test that the single dir structure includes 4 unique keys
     assert len(set(single_dir_result.keys())) == 4
+
+def test_avoid_na_row_output(
+    load_parsl_threaded: None, fx_tempdir: str, data_dir_cellprofiler: str
+):
+    """
+    Test to help detect and avoid scenarios where rows of NA-based data are returned
+    due to undetected objects in compartments (and as a result, imagenumbers).
+    For example, if there are imagenumbers in the image table and not the compartment table,
+    we want to avoid returning rows of NA data from the compartment tables after joins take place.
+    """
+
+    # run convert using a dataset known to contain the scenario outlined above.
+    parquet_file = convert(
+        source_path=(
+            f"{data_dir_cellprofiler}"
+            "/nf1_cellpainting_data/test-Plate_3_nf1_analysis.sqlite"
+        ),
+        dest_path=f"{fx_tempdir}/nf1_cellpainting_data/test-Plate_3_nf1_analysis.parquet",
+        dest_datatype="parquet",
+        preset="cellprofiler_sqlite_pycytominer",
+    )
+
+    # check that we have no nulls within Metadata_ImageNumber column
+    assert not pc.sum(
+        pc.is_null(
+            parquet.read_table(
+                source=parquet_file,
+            ).column("Metadata_ImageNumber")
+        )
+    ).as_py()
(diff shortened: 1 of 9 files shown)


tests/test_convert_threaded.py
```diff
@@ -237,3 +238,34 @@ def test_get_source_filepaths(
...
+
+
+def test_avoid_na_row_output(
+    load_parsl_threaded: None, fx_tempdir: str, data_dir_cellprofiler: str
+):
+    """
+    Test to help detect and avoid scenarios where rows of NA-based data are returned
+    due to undetected objects in compartments (and as a result, imagenumbers).
+    For example, if there are imagenumbers in the image table and not the compartment table,
+    we want to avoid returning rows of NA data from the compartment tables after joins take place.
```
gwaybio, line 250: I'm not sure I fully understand this comment. Is the imagenumber example you describe the only example? What other undetected objects could there be in a compartment? By definition, wouldn't we not have any knowledge of these undetected objects and therefore they wouldn't be included in the output parquet file at all?

gwaybio (approved):
Elegant fix, and I appreciate the added test and citation! One clarification question that need not delay merging

tests/test_convert_threaded.py
d33bs: Thanks for this comment - I think I incorrectly worded this, I'll correct it with the following:

`Test to help detect and avoid scenarios where rows of NA-based data are returned when using data sources where no compartment objects being detected in images from the data source (and as a result, compartment tables are missing imagenumbers when compared with the image table).`
    
and

`For example, if there are imagenumbers in the image table and not the compartment table,
    we want to avoid returning rows which include image table metadata and NA compartment feature data based on the way the join is composed. We advise using the compartment tables as the basis of data joins to generally avoid this issue, otherwise SQL processing will include data from the image table.`
    
When working through discovery and remediation of this issue with @jenna-tomkinson we noticed that the imagenumbers in the results which had NA compartment feature data from CytoTable output were often associated with images that had few-to-none objects or objects which were blurry / might not be interpreted by CellProfiler as valid objects.

gwaybio: Gotcha! Do these edits further clarify, simplify, and are accurate?

> Test to help detect and avoid scenarios where CytoTable returns rows of NA-based data. This occurs when CytoTable processes CellProfiler data sources with images that do not contain segmented objects. Thus, as a result, the CellProfiler image table contains imagenumbers that do not exist in any compartment object.

> Therefore, CytoTable does not return single-cell rows which include image table metadata and NA feature data. Using compartment tables as the basis of data joins avoids this issue.

d33bs: Accurate, thanks! I'll use this.

d33bs:
Thank you @gwaybio ! Merging this in.

Issue #187, opened 2024-04-12
Verify how large floating-point numbers are exported


d33bs:
While working on `cytominer-database` data comparisons for #30 I noticed variations in how floating point numbers are treated. As an example, the number `3.5215257120407011` may be found within `.tests/data/cytominer-database/data_b/A01-2/Cytoplasm.csv`. This number is extracted as `3.521525712040701` (one decimal place less, through automatic CSV settings in DuckDB interpreted as a DOUBLE). Similarly, `cyotminer-database` appears to change this number to `3.5215257120407006` (three additional decimal places, I believe this is done through Pandas CSV and/or `pd.Series.as_type()` casts).

I found that DuckDB can correctly interpret this number automatically as a [DECIMAL](https://duckdb.org/docs/sql/data_types/numeric.html#fixed-point-decimals) but that this option doesn't appear to be available through the CSV reader yet (planning to open an issue here). As it stands, this number is interpreted as a [DOUBLE](https://duckdb.org/docs/sql/data_types/numeric#floating-point-types), which truncates the data.

I did some work to try and research how this number is interpreted in Python, NumPy, Pandas, DuckDB, and Parquet. Both Python (through Decimal) and NumPy (through [longdouble](https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.longdouble)) were able to interpret an extended version of the number but not the precise number itself (similar to how `cytominer-database` operated above, adding decimal places). Pandas is able to accurately interpret the number through PyArrow types (specifically [decimal128](https://arrow.apache.org/docs/python/generated/pyarrow.decimal128.html#pyarrow.decimal128)). DuckDB seems to follow the same decimal-style formatting as PyArrow and is able to inference the width + scale when reading the number alone. From an Arrow decimal128 type, Parquet is able to write and read the number accurately through I believe the [decimal logical type](https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#decimal).

See here for a [Google Colab notebook](https://colab.research.google.com/drive/112x55m_YOfid3_Lt8SBdnbShCmoTl76l) with findings (and a [gist backup](https://gist.github.com/d33bs/bf6e71c381e1d8bcc50c4c69e500caf8)).

d33bs:
Opened https://github.com/duckdb/duckdb/issues/11639 in reference to this issue.

d33bs:
[...]

PR #188, opened 2024-04-12
Add tablenumber capabilities to distinguish repeated image numbers from multiple datasets

### LINKED ISSUE #30: How does CytoTable handle duplicate ImageNumbers?
`ImageNumber` is guaranteed to be unique only *within* an `Image.csv` but not across all `Image.csv` files in a dataset. 

Does `pycytominer-transform` handle this correctly? 

In [`cytominer-database`](https://github.com/cytomining/cytominer-database), we create a column, `TableNumber`, to identify rows in the `Image` table uniquely. `TableNumber` is set to be the hash of the `Image.csv` file:

https://github.com/cytomining/cytominer-database/blob/5aa00f58e4a31bbbd2a3779c87e7a3620b0030db/cytominer_database/ingest_variable_engine.py#L99

d33bs:
# Description

This PR addresses #30 by adding a tablenumber by user specification or automatically when there are multiple datasets being parsed by CytoTable. I encountered #187 during implementation, which I feel is a complex issue we shouldn't solve in this PR.

Closes #30 

# Checklist


cytotable/convert.py
```diff
@@ -173,6 +173,100 @@ def _prep_cast_column_data_types(
     return columns
 
 
+@python_app
+def _set_tablenumber(
+    sources: Dict[str, List[Dict[str, Any]]],
+    add_tablenumber: Optional[bool] = None,
+) -> Dict[str, List[Dict[str, Any]]]:
+    """
+    Gathers a "TableNumber" for the table which is a unique identifier intended
```
gwaybio, line 182: for which table?

cytotable/convert.py
```diff
@@ -173,6 +173,100 @@ def _prep_cast_column_data_types(
...
+    to help differentiate between imagenumbers to create distinct results.
```
gwaybio, line 183: create distinct results of what?

(i think it will be helpful to add a couple more words of specificity here)

cytotable/convert.py
```diff
@@ -173,6 +173,100 @@ def _prep_cast_column_data_types(
...
+        add_tablenumber: Optional[bool]
+            Whether to add a calculated tablenumber.
+            Note: when False, adds None as the tablenumber
+
+    Returns:
+        List[Dict[str, Any]]
+            New source group with added TableNumber details.
+    """
+
+    from cloudpathlib import AnyPath
```
gwaybio, line 201: ah! This import scares me - i thought that we removed cloudpathlib as a dependency?

cytotable/convert.py
```diff
@@ -173,6 +173,100 @@ def _prep_cast_column_data_types(
...
+        # create a data structure with the common parent for each dataset
+        # and the calculated checksum from the image table.
+        # note: the source_path parent is used for non-SQLite files
+        # whereas the direct source path is used for SQLite files.
+        (
+            str(source["source_path"].parent)
+            if source["source_path"].suffix != "sqlite"
+            else source["source_path"]
+        ): source["source_path"]
+        for source_group_name, source_group_vals in sources.items()
```
gwaybio, line 215:
```suggestion

        for source_group_name, source_group_vals in sources.items()
```

Would it be more clear to add a new line here? the dictionary comprehension is tripping me up a bit

cytotable/convert.py
```diff
@@ -310,15 +404,26 @@ def _source_pageset_to_parquet(
...
     pathlib.Path(source_dest_path).mkdir(parents=True, exist_ok=True)
 
+    # build tablenumber segment addition (if necessary)
+    tablenumber_sql = (
+        # to become tablenumber in sql select later with bigint (8-byte integer)
+        # we cast here to bigint to avoid concat or join conflicts later due to
+        # misaligned automatic data typing.
+        f"CAST({source['tablenumber']} AS BIGINT) as TableNumber, "
+        if source["tablenumber"] is not None
+        # if we don't have a tablenumber value, don't introduce the column
```
gwaybio, line 414:
```suggestion
        # if we already have a tablenumber value, don't introduce the column
```

Am I reading this correctly?

tests/conftest.py
```diff
@@ -214,6 +214,98 @@ def cytominerdatabase_to_pycytominer_merge_single_cells_parquet(
     return output_paths
 
 
+@pytest.fixture()
+def cytominerdatabase_to_manual_join_parquet(
```
gwaybio, line 218: great to see these tests!

gwaybio (approved):

cytotable/convert.py
d33bs: Thanks good question - I'll add some notes on this.

cytotable/convert.py
d33bs: Thanks, good point - I'll add some notes here.

cytotable/convert.py
d33bs: Thanks, yes we did talk about removing it but decided to leave it intact and update it (I think). So far, it's been working great in the context of JUMP (see https://github.com/WayScience/JUMP-single-cell/pull/33), where we use that capability to help interact with S3 based files without added code complexity.

cytotable/convert.py
d33bs: Thanks yes, I'll add this in!

cytotable/convert.py
d33bs: Thanks for checking on this - I should be more clear here, adding tablenumber values is optional via a parameter to CytoTable. If we aren't supposed to add tablenumbers we evaluate this to an empty string (meaning we won't add a tablenumber).

tests/conftest.py
d33bs: Cheers, agreed!

d33bs:
Thanks for this review @gwaybio ! I'm going to merge this in after responding to your comments.

Issue #211, opened 2024-06-25
Possible Parsl manager persistence after CytoTable processing


d33bs:
After running CytoTable it seems that sometimes the Parsl manager persists for potential reuse without shutting down. Requires further investigation.

PR #212, opened 2024-06-25
Use real AWS S3 data tests and apply related fixes

### LINKED ISSUE #193: Add large test data cases to help detect non-deterministic challenges
> Probably best served by a different PR, but I'm wondering if it is worth adding a test that ensures, in a large enough dataset, that there are no duplicate rows and all expected rows are present.

_Originally posted by @gwaybio in https://github.com/cytomining/CytoTable/pull/182#pullrequestreview-2007415041_

This could take place with a synthetic dataset which is created through duplication of source data rows with minor variations. It could also reference a remote dataset.

Likely useful to add a Pytest marker indicating the nature of the test as "large" to allow development to skip the test by default.

### LINKED ISSUE #198: Error on processing `moto`-based files within scheduled, un-locked dependency tests
An error occurred with two of the tests during a scheduled and un-locked dependency test earlier today. The only failing test in these cases is: `tests/test_convert_threaded.py::test_convert_s3_path_csv`. This seems to happen due to a missing file from the mocked `moto` bucket to a local file, resulting in a failed data query from DuckDB.

Reference to the job: https://github.com/cytomining/CytoTable/actions/runs/8812487779/job/24204265980

```console
duckdb.duckdb.InvalidInputException: Invalid Input Error: Attempting to execute an unsuccessful or closed pending query result
Error: IO Error: No files found that match the pattern "/tmp/tmpv8ak2gtt/example/2/cells.csv"
...
FAILED tests/test_convert_threaded.py::test_convert_s3_path_csv - parsl.dataflow.errors.DependencyError: <exception str() failed>
```

d33bs:
# Description

This changes in this PR address #198 by removing `moto` and related tests to avoid existing and future challenges with mocked `s3` resources (in addition to the failing tests, a short justification can be [found here](https://github.com/cytomining/CytoTable/issues/198#issuecomment-2152947239)). Moving forward, CytoTable will now be tested on a CSV and SQLite resource from the [`cellpainting-gallery`](https://registry.opendata.aws/cellpainting-gallery/) (as outlined in the fixtures). As a result, I believe we now are addressing #193 within reason because of the SQLite addition (please don't hesitate to let me know if you feel otherwise and we should keep that issue open).

In the process of developing towards this fix I added a new preset which enables compatibility with JUMP data (`cpg0016-jump`) from the `cellpainting-gallery` in order to effectively perform a test from an S3 SQlite object (no other presets appeared to exactly match this need). CC @jenna-tomkinson 

Some notes:
- I noticed functions within the `sources` module experiencing challenges from Parsl app decoration due to how the `cloudpathlib` client is constructed and how logging operates with HTEX executor tasks (which sometimes makes errors less visible). Thinking through this had me recognize that there didn't seem to be a major benefit from these being decorated this way. Because of this, I changed these functions to be un-decorated, which simplified troubleshooting and feels like a good option to reduce complexity in the future.
- I moved logic found within `convert` for checking that a file has at least one row to `sources` through `_file_is_more_than_one_line` to help keep the use of `cloudpathlib.CloudPath`'s consistent and filter data sources earlier in the flow before we begin processing them.
- I noticed `cytominer-database` was consistently experiencing challenges within Python 3.12 environments (tests couldn't find the `ingest` command line option). As a result, I've moved to static files produced from `cytominer-database` which increase the size of this repository but should help us save time during tests and avoid the error I was seeing during development.
- After removing the `cytominer-database` processing step, I noticed that there were some missing CSV's from the `cytominer-database` test data, which have been added in this PR.
- A chunk size of `30000` was used after trying many different iterations with the S3 SQLite file test. GitHub Actions runner containers appear to experience resource challenges at sizes above this. GitHub Actions doesn't appear to provide exact details on resource use overages upfront, so I expect the chunk size may need tinkering over time.
- Sorting results for any chunk size with the S3 SQLite file appeared to trigger GitHub Actions runner container challenges so I've turned this off for that particular test.
- I've split the `large_data_tests` marked tests to be run as a separate job. Running things together with the other tests and test matrix appeared to have trouble with resources available in GitHub Actions runner containers. While the time involved to run the single test in this low for the time being, we may need to split it into another scheduled test (or something similar) if/when the mark's tests grow. This may be related to challenges with #211, Python garbage collection, or GitHub Actions runner container storage availability (more investigation is needed).
- A minor fix to address an issue related to https://github.com/duckdb/duckdb/issues/3243 was applied to `duckdb` connection setup (this seems to sporadically appear, disappear, and reappear again within GitHub Actions runner environments).

Closes #198 
Closes #193 

# Checklist


.github/workflows/test.yml
```diff
@@ -24,13 +24,13 @@ jobs:
...
       - name: Python setup
-        uses: actions/setup-python@v4
+        uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python_version }}
       # remove poetry.lock file for scheduled tests
       # to help simulate possible upstream issues
       - name: Remove poetry.lock for scheduled tests
-        if: github.event_name == 'schedule'
+        if: github.event.schedule == '0 7 * * 3'
```
gwaybio, line 33: It might be worth adding a bit of detail in a comment specifying what this means exactly

tests/conftest.py
```diff
@@ -792,3 +696,28 @@ def example_sqlite_mixed_types_database(
     finally:
         # after completing the tests, remove the file
         pathlib.Path(filepath).unlink()
+
+
+@pytest.fixture(name="example_s3_path_csv_jump")
+def fixture_example_s3_path_csv_jump() -> str:
+    """
+    Provides an example s3 endpoint for use with tests
```
gwaybio, line 704: It's a bit unclear to me why this isn't pointing to a csv file? Maybe you can clarify slightly in the docstring?

tests/test_convert_threaded.py
```diff
@@ -125,72 +87,28 @@ def test_convert_s3_path_sqlite(
...
-    )
-    assert local_cytotable_table.sort_by(
-        [(name, "ascending") for name in local_cytotable_table.schema.names]
-    ).equals(
-        s3_cytotable_table_nested.sort_by(
-            [(name, "ascending") for name in s3_cytotable_table_nested.schema.names]
-        )
-    )
+    # check the shape of the data
+    assert (parquet_file_meta.num_rows, parquet_file_meta.num_columns) == (74226, 5928)
```
gwaybio, line 111: It doesn't look like you're checking for any sorting issue here? Maybe this is good b/c it is focused, but I wonder if testing here would be good? I really don't know if this is a good suggestion!

gwaybio (approved):
LGTM! I went in thinking the review would be brutal, but I think it is relatively bite-sized. A few comments

.github/workflows/test.yml
d33bs: Thanks for this comment! I made this change as initially I felt that the test might be so large that it'd warrant an additional schedule (in addition to the Wednesday scheduled runs). This was a way to specify exactly which schedule was used as a trigger for a task within the job when there might be multiple. I retained it because it felt more declarative and we could end up with more scheduled tests yet in the future. I'll add a comment as you suggested to keep this well documented (`crontab` on its own is a bit terse to read!).

tests/conftest.py
d33bs: Thanks good point! This points to a directory with CSV files within object storage. We point to the directory to help include all compartments and metadata CSV files for processing within CytoTable.

tests/test_convert_threaded.py
d33bs: Thanks, this is a great question! After thinking about this and seeing issues in other testing without sorts I moved to retain sorted output (the default) by removing `sort_output=False` and reduced the number of columns needed to perform a join. I reduced the columns to help avoid memory issues I was seeing in GitHub Actions runners. This said, I feel that addressing the join sort memory consumption should be a new issue / PR to help reduce the overall resource requirements for CytoTable.

d33bs:
Thank you @gwaybio ! After making some changes I feel good about how this now looks. I plan to create a new issue to explore memory resource reductions for sorted joins.

PR #220, opened 2024-08-01
Move to keyset pagination

### LINKED ISSUE #214: Reduce sorting memory resource requirements
When using `join=True` and a data source which is larger than a gigabyte `sort_output=True` appears to cause a large memory spike during join operations. This may cause a system crash in scenarios where not enough memory is available.

d33bs:
# Description

This PR modifies "chunk" behavior to follow keyset pagination patterns. We gather sets of pages as `pagesets` which are based on a key per table which is used to help create pages of rows from tables. `pagesets` are gathered ahead of time to retain the ability to parallelize where appropriate. This change required a new configuration parameter to be specified per table and the overall join.

Operationally this reduces the amount of data stored in tables throughout, reduces the complexity of the sorting performed, and avoids unnecessary data loading which occurs during the use of `OFFSET`.

Changes here made it unnecessary to create metadata columns and as a result I've removed those from all spots I could find.

Changes here are intended to address #214 , which turned out to be more related to data size than the sorting implementation. This is confirmed by a now passing `large_data_tests` test of JUMP data using the full dataset (instead of a truncated version).

Closes #214

# Checklist

### DIFF
--- a/cytotable/constants.py
+++ b/cytotable/constants.py
@@ -68,13 +68,6 @@ SQLITE_AFFINITY_DATA_TYPE_SYNONYMS = {
     ],
 }
 
-# metadata column names and types for internal use within CytoTable
-CYOTABLE_META_COLUMN_TYPES = {
-    "cytotable_meta_source_path": "VARCHAR",
-    "cytotable_meta_offset": "BIGINT",
-    "cytotable_meta_rownum": "BIGINT",
-}
-
 CYTOTABLE_DEFAULT_PARQUET_METADATA = {
     "data-producer": "https://github.com/cytomining/CytoTable",
     "data-producer-version": str(_get_cytotable_version()),
--- a/cytotable/convert.py
+++ b/cytotable/convert.py
@@ -86,7 +86,7 @@ def _get_table_columns_and_types(
     # with exception handling to read mixed-type data
     # using sqlite3 and special utility function
     try:
-        # isolate using new connection to read data with chunk size + offset
+        # isolate using new connection to read data based on pageset
         # and export directly to parquet via duckdb (avoiding need to return data to python)
         # perform the query and create a list of dictionaries with the column data for table
         with _duckdb_reader() as ddb_reader:
@@ -283,7 +283,6 @@ def _get_table_keyset_pagination_sets(
 def _source_chunk_to_parquet(
     source_group_name: str,
     source: Dict[str, Any],
-    chunk_size: int,
     pageset: Tuple[int, int],
     dest_path: str,
     sort_output: bool,
@@ -297,10 +296,8 @@ def _source_chunk_to_parquet(
         source: Dict[str, Any]
             Contains the source data to be chunked. Represents a single
             file or table of some kind along with collected information about table.
-        chunk_size: int
-            Row count to use for chunked output.
-        page_keys: Dict[str, str]
-            ...
+        page_keys: str:
+            The table and column names to be used for key pagination.
         pageset: Tuple[int, int]
             The pageset for chunking the data from source.
         dest_path: str
@@ -318,7 +315,6 @@ def _source_chunk_to_parquet(
     import duckdb
     from cloudpathlib import AnyPath
 
-    from cytotable.constants import CYOTABLE_META_COLUMN_TYPES
     from cytotable.utils import (
         _duckdb_reader,
         _sqlite_mixed_type_query_to_parquet,
@@ -331,12 +327,6 @@ def _source_chunk_to_parquet(
         f"{str(source['source_path'].parent.name).lower()}"
     )
     pathlib.Path(source_dest_path).mkdir(parents=True, exist_ok=True)
-
-    source_path_str = (
-        source["source_path"]
-        if "table_name" not in source.keys()
-        else f"{source['source_path']}_table_{source['table_name']}"
-    )
 
     # add source table columns
     casted_source_cols = [
@@ -451,10 +441,7 @@ def _prepend_column_name(
 
     import pyarrow.parquet as parquet
 
-    from cytotable.constants import (
-        CYOTABLE_META_COLUMN_TYPES,
-        CYTOTABLE_ARROW_USE_MEMORY_MAPPING,
-    )
+    from cytotable.constants import CYTOTABLE_ARROW_USE_MEMORY_MAPPING
     from cytotable.utils import _write_parquet_table_with_metadata
 
     logger = logging.getLogger(__name__)
@@ -502,10 +489,8 @@ def _prepend_column_name(
         #   source_group_name_stem: 'Cells'
         #   column_name: 'AreaShape_Area'
         #   updated_column_name: 'Cells_AreaShape_Area'
-        if (
-            column_name not in identifying_columns
-            and not column_name.startswith(source_group_name_stem.capitalize())
-            and column_name not in CYOTABLE_META_COLUMN_TYPES
+        if column_name not in identifying_columns and not column_name.startswith(
+            source_group_name_stem.capitalize()
         ):
             updated_column_names.append(f"{source_group_name_stem}_{column_name}")
         # if-condition for prepending 'Metadata_' to column name
@@ -630,7 +615,6 @@ def _concat_source_group(
         CYTOTABLE_DEFAULT_PARQUET_METADATA,
     )
     from cytotable.exceptions import SchemaException
-    from cytotable.utils import _write_parquet_table_with_metadata
 
     # build a result placeholder
     concatted: List[Dict[str, Any]] = [
@@ -868,7 +852,6 @@ def _concat_join_sources(
         CYTOTABLE_ARROW_USE_MEMORY_MAPPING,
         CYTOTABLE_DEFAULT_PARQUET_METADATA,
     )
-    from cytotable.utils import _write_parquet_table_with_metadata
 
     # remove the unjoined concatted compartments to prepare final dest_path usage
     # (we now have joined results)
@@ -1121,11 +1104,11 @@ def _to_parquet(  # pylint: disable=too-many-arguments, too-many-locals
         for source_group_name, source_group_vals in sources.items()
     }
 
-    # if offsets is none and we haven't halted, remove the file as there
+    # if pagesets is none and we haven't halted, remove the file as there
     # were input formatting errors which will create challenges downstream
     invalid_files_dropped = {
         source_group_name: [
-            # ensure we have offsets
+            # ensure we have pagesets
             source
             for source in source_group_vals
             if source["pagesets"] is not None
@@ -1166,11 +1149,10 @@ def _to_parquet(  # pylint: disable=too-many-arguments, too-many-locals
                     "table": [
                         # perform column renaming and create potential return result
                         _prepend_column_name(
-                            # perform chunked data export to parquet using offsets
+                            # perform chunked data export to parquet using pagesets
                             table_path=_source_chunk_to_parquet(
                                 source_group_name=source_group_name,
                                 source=source,
-                                chunk_size=chunk_size,
                                 pageset=pageset,
                                 dest_path=expanded_dest_path,
                                 sort_output=sort_output,
@@ -1344,6 +1326,8 @@ def convert(  # pylint: disable=too-many-arguments,too-many-locals
             A dictionary mapping data type groups to specific types.
             Roughly includes Arrow data types language from:
             https://arrow.apache.org/docs/python/api/datatypes.html
+        page_keys: str:
+            The table and column names to be used for key pagination.
         sort_output: bool (Default value = True)
             Specifies whether to sort cytotable output or not.
         drop_null: bool (Default value = False)
--- a/cytotable/convert.py
+++ b/cytotable/convert.py
@@ -180,7 +180,7 @@ def _get_table_keyset_pagination_sets(
     page_key: str,
     source: Optional[Dict[str, Any]] = None,
     sql_stmt: Optional[str] = None,
-) -> Union[List[Any], None]:
+) -> Union[List[Tuple[Union[int, float], Union[int, float]]], None]:
     """
     Get table data chunk keys for later use in capturing segments
     of values. This work also provides a chance to catch problematic
@@ -261,7 +261,7 @@ def _get_table_keyset_pagination_sets(
 def _source_chunk_to_parquet(
     source_group_name: str,
     source: Dict[str, Any],
-    pageset: Tuple[int, int],
+    pageset: Tuple[Union[int, float], Union[int, float]],
     dest_path: str,
     sort_output: bool,
 ) -> str:
@@ -274,8 +274,6 @@ def _source_chunk_to_parquet(
         source: Dict[str, Any]
             Contains the source data to be chunked. Represents a single
             file or table of some kind along with collected information about table.
-        page_keys: str:
-            The table and column names to be used for key pagination.
         pageset: Tuple[int, int]
             The pageset for chunking the data from source.
         dest_path: str
--- a/cytotable/convert.py
+++ b/cytotable/convert.py
@@ -43,10 +43,7 @@ def _get_table_columns_and_types(
             list of dictionaries which each include column level information
     """
 
-    import pathlib
-
     import duckdb
-    from cloudpathlib import AnyPath
 
     from cytotable.utils import _duckdb_reader, _sqlite_mixed_type_query_to_parquet
 
@@ -102,6 +99,7 @@ def _get_table_columns_and_types(
             )
 
     except duckdb.Error as e:
+        print(source["pagesets"])
         # if we see a mismatched type error
         # run a more nuanced query through sqlite
         # to handle the mixed types
@@ -206,7 +204,8 @@ def _get_table_keyset_pagination_sets(
     import logging
 
     import duckdb
-    from cloudpathlib import AnyPath, CloudPath
+    import sqlite3
+    from contextlib import closing
 
     from cytotable.exceptions import NoInputDataException
     from cytotable.utils import _duckdb_reader
@@ -225,8 +224,20 @@ def _get_table_keyset_pagination_sets(
                 else:
                     sql_query = f"SELECT {page_key} FROM sqlite_scan('{source_path}', '{table_name}') ORDER BY {page_key}"
 
-                keys = ddb_reader.execute(sql_query).fetchall()
-                keys = [key[0] for key in keys]
+                keys = [key[0] for key in ddb_reader.execute(sql_query).fetchall()]
+
+        # exception case for when we have mixed types
+        # (i.e. integer col with string and ints) in a sqlite column
+        except duckdb.TypeMismatchException:
+            with closing(sqlite3.connect(source_path)) as cx:
+                with cx:
+                    keys = [
+                        key[0]
+                        for key in cx.execute(
+                            f"SELECT {page_key} FROM {table_name} ORDER BY {page_key};"
+                        ).fetchall()
+                        if isinstance(key[0], (int, float))
+                    ]
 
         except (
             duckdb.InvalidInputException,
@@ -245,10 +256,25 @@ def _get_table_keyset_pagination_sets(
             keys = [key[0] for key in keys]
 
     # Create chunks of keys
-    chunks = [
-        ((values := keys[i : i + chunk_size])[0], values[-1])
-        for i in range(0, len(keys), chunk_size)
-    ]
+    chunks = []
+    i = 0
+    while i < len(keys):
+        start_key = keys[i]
+        end_index = min(i + chunk_size, len(keys)) - 1
+        end_key = keys[end_index]
+
+        # Ensure non-overlapping by incrementing the start of the next range
+        next_start_index = end_index + 1
+        if next_start_index < len(keys):
+            next_start_key = keys[next_start_index]
+            while next_start_key == end_key and next_start_index + 1 < len(keys):
+                next_start_index += 1
+                next_start_key = keys[next_start_index]
+            chunks.append((start_key, next_start_key - 1))
+        else:
+            chunks.append((start_key, end_key))
+
+        i = next_start_index
 
     return chunks
 
@@ -687,7 +713,6 @@ def _concat_source_group(
 def _prepare_join_sql(
     sources: Dict[str, List[Dict[str, Any]]],
     joins: str,
-    sort_output: bool,
 ) -> str:
     """
     Prepare join SQL statement with actual locations of data based on the sources.
@@ -727,7 +752,6 @@ def _prepare_join_sql(
 def _join_source_chunk(
     dest_path: str,
     joins: str,
-    chunk_size: int,
     page_key: str,
     pageset: Tuple[int, int],
     sort_output: bool,
@@ -1062,6 +1086,16 @@ def _to_parquet(  # pylint: disable=too-many-arguments, too-many-locals
     # expand the destination path
     expanded_dest_path = _expand_path(path=dest_path)
 
+    # check that each source group name has a pagination key
+    for source_group_name in sources.keys():
+        matching_keys = [
+            key for key in page_keys.keys() if key.lower() in source_group_name.lower()
+        ]
+        if not matching_keys:
+            raise CytoTableException(
+                f"No matching key found in page_keys for source_group_name: {source_group_name}"
+            )
+
     # prepare pagesets for chunked data export from source tables
     pagesets_prepared = {
         source_group_name: [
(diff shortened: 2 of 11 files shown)


cytotable/convert.py
```diff
@@ -285,10 +295,10 @@ def _source_chunk_to_parquet(
         source: Dict[str, Any]
             Contains the source data to be chunked. Represents a single
             file or table of some kind along with collected information about table.
-        chunk_size: int
-            Row count to use for chunked output.
-        offset: int
-            The offset for chunking the data from source.
+        page_keys: str:
+            The table and column names to be used for key pagination.
```
falquaddoomi, line 299: I don't see `page_keys` in the list of arguments to this function or any references to it in the body.

cytotable/convert.py
```diff
@@ -245,34 +248,41 @@ def _get_table_chunk_offsets(
...
+    return chunks
 
 
 @python_app
 def _source_chunk_to_parquet(
     source_group_name: str,
     source: Dict[str, Any],
-    chunk_size: int,
-    offset: int,
+    pageset: Tuple[int, int],
```
falquaddoomi, line 285: I see that the type here is `Tuple[int, int]`; does that imply that columns identified in `page_keys` are also integers? If so, you might want to add that to the documentation and/or add a check that they are in fact integers.

cytotable/convert.py
```diff
@@ -183,13 +175,14 @@ def _prep_cast_column_data_types(
 
 
 @python_app
-def _get_table_chunk_offsets(
+def _get_table_keyset_pagination_sets(
```
falquaddoomi, line 178: It seems that `page_key` needs to be orderable and from the definition of `pagesets` elsewhere I presume it also has to be an integer, but is it the case that it also has to be unique? If there are repeats, I imagine you'd end up with page sizes larger than the chunk size you specified. Perfectly fine if the page key needs to be unique, but that's something I'd add to the documentation if so.

You might also consider tightening the return type from `List[Any]` to `List[Tuple[int,int]]` to be more explicit about the chunks being a list of integer key pairs that define each page.

cytotable/convert.py
```diff
@@ -245,34 +248,41 @@ def _get_table_chunk_offsets(
...
-    return list(
-        range(
-            0,
-            # gather rowcount from table and use as maximum for range
-            rowcount,
-            # step through using chunk size
-            chunk_size,
-        )
-    )
+        i = next_start_index
```
falquaddoomi, line 276: I think this code is fine as-is, but for brevity you might consider using the `grouper()` function as described in the [itertools docs](https://docs.python.org/3.10/library/itertools.html) -- unfortunately I can't link to the function directly, but you'll find it if you search for "def grouper".

A simplified version of that function, [`chunked()`](https://more-itertools.readthedocs.io/en/stable/api.html#more_itertools.chunked), is included in the `more-itertools` library, but I don't know if you want to take an extra dependency for such a simple function.

You could use the `chunked()` function like so to produce your set of non-overlapping intervals:
```python
chunks = [(x[0], x[-1]) for x in chunked(keys, chunk_size)]
```

If you instead use `grouper()`, you'll have to account for the case in which a group ends up padded with `None` if it can't construct a group of size `chunk_size`.
```python
[(x[0], x[-1] if x[-1] is not None else x[0]) for x in grouper(keys, chunk_size)]
```

It seems that SQL's `BETWEEN` is inclusive, so a page like `(100, 100)` will still match the keys with value `100`.

falquaddoomi (commented):
Neat solution to the problem you were facing; I agree that it's less brittle than using `LIMIT` and `OFFSET` since you're now relying on the data rather than implementation-specific definitions of `LIMIT` and `OFFSET`.

I think you could add a bit more description of what columns are good candidates for keys; from my reading of your code I assume they'd be unique integers.

Anyway, nice work, and glad to see this issue put to rest!

cytotable/convert.py
d33bs: Thanks great catch, I'll update accordingly!

cytotable/convert.py
d33bs: Thanks, good point - as designed `page_keys` need to be numeric (ints or floats). I'll update the type here along with the documentation aspects.

cytotable/convert.py
d33bs: Thanks good point! Yes, there are some assumptions we should document upfront for the `pagesets`. I tried to be loose in the considerations here but there are a few gotchas which resulted in a more complicated implementation of the `pageset` list of tuples implementation (will touch on this in the other comment).

cytotable/convert.py
d33bs: Thanks I'll explore this a bit more! I found that `grouper()` was challenging to use with concerns to overlaps that should respect the chunk size. There's likely a more elegant solution as you suggested.

d33bs: Thanks again. I reworked this to simplify and demonstrate reasoning. I decided to create a function out of this work and test it individually to ensure we know what we're receiving (and why the implementation doesn't use existing algorithms in built-in's). My hope here is that the test cases help demonstrate better what output we're expecting (it's a bit different than `grouper`, `groupby`, or `more_itertools.chunked` provide based on repeated value and SQL-specific expectations).

d33bs:
Thanks @falquaddoomi for the review and comments! I made some adjustments based on what you mentioned.

cytotable/convert.py
```diff
@@ -199,21 +192,27 @@ def _get_table_chunk_offsets(
...
-            List of integers which represent offsets to use for reading
-            the data later on.
+        List[Any]
+            List of keys to use for reading the data later on.
     """
 
     import logging
-    import pathlib
+    import sqlite3
+    from contextlib import closing
```
jenna-tomkinson, line 210: Why do imports happen at two points in this file? I could be mistaken, but isn't it more standard to have all imports at the start of the file?

cytotable/convert.py
```diff
@@ -43,10 +43,7 @@ def _get_table_columns_and_types(
             list of dictionaries which each include column level information
     """
 
-    import pathlib
-
```
jenna-tomkinson, line 47: Why is pathlib removed?

cytotable/convert.py
```diff
@@ -245,34 +253,20 @@ def _get_table_chunk_offsets(
...
-            rowcount,
-            # step through using chunk size
-            chunk_size,
-        )
-    )
+    return _generate_pagesets(keys, chunk_size)
 
 
 @python_app
 def _source_chunk_to_parquet(
```
jenna-tomkinson, line 266: Should this function be renamed to not use chunk since based on what I can see, you don't use chunk_size anymore? Would `_source_pageset_to_parquet` be more accurate?

cytotable/convert.py
```diff
@@ -349,7 +319,7 @@ def _source_chunk_to_parquet(
     # create selection statement from lists above
     select_columns = ",".join(
         # if we should sort the output, add the metadata_cols
-        cytotable_metadata_cols + casted_source_cols
+        casted_source_cols
```
jenna-tomkinson, line 322: How are metadata columns added now with this change?

tests/test_convert_threaded.py
```diff
@@ -102,41 +102,16 @@ def test_convert_s3_path_sqlite_join(
...
-                (SELECT * FROM read_parquet('cytoplasm.parquet') LIMIT 5000) AS cytoplasm
-            LEFT JOIN (SELECT * FROM read_parquet('cells.parquet') LIMIT 5000) AS cells ON
-                cells.Metadata_ImageNumber = cytoplasm.Metadata_ImageNumber
-                AND cells.Metadata_ObjectNumber = cytoplasm.Cytoplasm_Parent_Cells
-            LEFT JOIN (SELECT * FROM read_parquet('nuclei.parquet') LIMIT 5000) AS nuclei ON
-                nuclei.Metadata_ImageNumber = cytoplasm.Metadata_ImageNumber
-                AND nuclei.Metadata_ObjectNumber = cytoplasm.Cytoplasm_Parent_Nuclei
-            LEFT JOIN (SELECT * FROM read_parquet('image.parquet') LIMIT 5000) AS image ON
-                image.Metadata_ImageNumber = cytoplasm.Metadata_ImageNumber
-        """,
```
jenna-tomkinson, line 129: Was this deleted because of the pagination fix?

cytotable/convert.py
falquaddoomi: That's generally a good point, but I believe this is because the parallelization library that @d33bs is using, [parsl](https://parsl-project.org/), requires that functions that are parallelized import all the libraries they need in the function, not within the enclosing module like you'd normally do. You can tell it's an entrypoint from the `@python_app` decorator before the function definition. 

Here's a bit from the parsl docs at least saying it's necessary to import within parallelized functions: https://parsl.readthedocs.io/en/stable/1-parsl-introduction.html#Python-Apps.

cytotable/constants.py
```diff
@@ -68,13 +68,6 @@
     ],
 }
 
-# metadata column names and types for internal use within CytoTable
-CYOTABLE_META_COLUMN_TYPES = {
-    "cytotable_meta_source_path": "VARCHAR",
-    "cytotable_meta_offset": "BIGINT",
-    "cytotable_meta_rownum": "BIGINT",
-}
```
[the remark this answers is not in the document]
d33bs: Thanks great question! These columns were added to help implement offset-based pagination (avoiding the use of existing table columns). I opted to remove them as keyset pagination focuses on using existing table keys for pagination. The removal also reduces the complexity of the code more generally which likely has performance improvements (we're adding 3 fewer columns to each table and each chunk of each table).

cytotable/convert.py
d33bs: Thanks Jenna for a good question and Faisal for the docs. Concurring with what Faisal mentioned, this is to help with multiprocessing through Parsl where each decorated function might not have access of the full module's imports.

cytotable/convert.py
d33bs: Thanks - `pathlib` was removed as I noticed it was no longer needed for `_get_table_columns_and_types`. I believe the import inadvertently remained after earlier changes than this PR, so this was a bit of opportunistic refactoring (I noticed it during the work towards this PR).

cytotable/convert.py
d33bs: Thanks, great point - I like that change and will make a further revision to avoid the use of "chunk" to defer to "page" where possible.

cytotable/convert.py
d33bs: Thanks for asking about this, CytoTable-specific metadata columns (columns which add CytoTable data) are no longer added to tables. Metadata columns from the source data (for ex. from CellProfiler) will remain unchanged in terms of how they are processed by CytoTable.

tests/test_convert_threaded.py
d33bs: Thanks - yes, this is removed to defer to the preset default SQL for joins. The change here is mostly to remove `LIMIT 5000` from each joined table (from Parquet datasets) and instead use the full dataset. With this change we are now able to process the full dataset within the scope of a [GitHub Actions runner image's resources](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories) (4 CPU, 16 GB RAM). Earlier changes towards offset-based pagination resulted in #214 and an inability to run the full data through this test. Keyset pagination implemented in this PR is much more performant and doesn't require us to artificially limit the query here anymore.

cytotable/convert.py
d33bs: Thinking about this more, I'm going to make minimal changes to the usage of those words. Generally I feel we need to distinguish between Parquet pages vs CytoTable pages, and retaining "chunk" might be more understandable for now (in addition to avoiding a change to the `convert` API.

cytotable/convert.py
```diff
@@ -223,18 +222,27 @@ def _get_table_chunk_offsets(
...
-                        if source_type == ".csv"
-                        else f"SELECT COUNT(*) from sqlite_scan('{source_path}', '{table_name}')"
-                    ).fetchone()[0]
-                )
+                if source_type == ".csv":
+                    sql_query = f"SELECT {page_key} FROM read_csv_auto('{source_path}', header=TRUE, delim=',') ORDER BY {page_key}"
+                else:
+                    sql_query = f"SELECT {page_key} FROM sqlite_scan('{source_path}', '{table_name}') ORDER BY {page_key}"
+
+                keys = [key[0] for key in ddb_reader.execute(sql_query).fetchall()]
```
gwaybio, line 231: is there a better name for this variable? maybe `page_keys`?

cytotable/convert.py
```diff
@@ -245,34 +253,20 @@ def _get_table_chunk_offsets(
...
-        # gather the total rowcount from csv or sqlite data input sources
         with _duckdb_reader() as ddb_reader:
-            rowcount = int(
-                ddb_reader.execute(
-                    # nosec
-                    f"SELECT COUNT(*) FROM ({sql_stmt})"
-                ).fetchone()[0]
-            )
+            sql_query = f"SELECT {page_key} FROM ({sql_stmt}) ORDER BY {page_key}"
+            keys = ddb_reader.execute(sql_query).fetchall()
```
gwaybio, line 259: same comment on variable name here

cytotable/convert.py
```diff
@@ -377,14 +347,9 @@ def _source_chunk_to_parquet(
...
+                    WHERE {source['page_key']} BETWEEN {pageset[0]} AND {pageset[1]}
                     /* order by all columns for deterministic output */
-                    ORDER BY ALL
-                    LIMIT {chunk_size} OFFSET {offset}
-                    """
-                    if sort_output
-                    else f"""
-                    {base_query}
-                    LIMIT {chunk_size} OFFSET {offset}
+                    {"ORDER BY " + source['page_key'] if sort_output else ""};
```
gwaybio, line 352: I'm likely not reading this correctly, but is this ordering by `page_key` for every chunked/pageset operation? Wouldn't it be more efficient to order after all pagesets are processed?

cytotable/convert.py
```diff
@@ -798,31 +745,19 @@ def _join_source_chunk(
...
-                /* exclude metadata columns from the results
-                by using a lambda on column names based on exclude_meta_cols. */
-                COLUMNS (c -> ({" AND ".join(exclude_meta_cols)}))
-                FROM joined;
-                """
+            )
+            SELECT *
+            FROM joined
+            WHERE {page_key} BETWEEN {pageset[0]} AND {pageset[1]}
+            {"ORDER BY " + page_key if sort_output else ""};
```
gwaybio, line 759: Same question about ordering

cytotable/convert.py
```diff
@@ -1112,34 +1050,52 @@ def _to_parquet(  # pylint: disable=too-many-arguments, too-many-locals
...
-    # prepare offsets for chunked data export from source tables
-    offsets_prepared = {
+    # check that each source group name has a pagination key
+    for source_group_name in sources.keys():
+        matching_keys = [
+            key for key in page_keys.keys() if key.lower() in source_group_name.lower()
+        ]
+        if not matching_keys:
+            raise CytoTableException(
+                f"No matching key found in page_keys for source_group_name: {source_group_name}"
```
gwaybio, line 1060: Can you add a bit more details to this exception? Any further details about how a user might be able to resolve this exception would be helpful. Why might someone encounter this error?

cytotable/convert.py
```diff
@@ -1440,6 +1408,21 @@ def convert(  # pylint: disable=too-many-arguments,too-many-locals
...
+            else page_keys
+        )
+
+    # Raise an exception for scenarios where one configures CytoTable to join
+    # but does not provide a pagination key for the joins.
+    if join and (page_keys is None or "join" not in page_keys.keys()):
+        raise CytoTableException(
+            (
+                "When using join=True one must pass a 'join' pagination key "
+                "in the page_keys parameter."
```
gwaybio, line 1423: Can you answer the question "why?" in this error message? Or perhaps provide an example?

cytotable/utils.py
```diff
@@ -188,10 +189,10 @@ def _sqlite_mixed_type_query_to_parquet(
             A str which is a path to a SQLite database file.
         table_name: str:
             The name of the table being queried.
-        chunk_size: int:
-            Row count to use for chunked output.
-        offset: int:
-            The offset for chunking the data from source.
+        page_key: str:
+            The column name to be used as the key for pagination.
```
gwaybio, line 193: Might be helpful to define the variable without using the variable name. Perhaps something like "The column name to be used to identify and join different pagination chunks"?

cytotable/utils.py
```diff
@@ -188,10 +189,10 @@ def _sqlite_mixed_type_query_to_parquet(
...
+        pageset: Tuple[int, int]:
+            The pageset for chunking the data from source.
```
gwaybio, line 195: Similar comment here - this is the chunk identifier for some indicator of where the chunk is in context with the rest of the data?

cytotable/utils.py
```diff
@@ -268,42 +266,14 @@ def _sqlite_affinity_data_type_lookup(col_type: str) -> str:
...
-            """
-            if sort_output
-            else f"""
+        sql_stmt = f"""
             SELECT
                 {', '.join(query_parts)}
             FROM {table_name}
-            LIMIT {chunk_size} OFFSET {offset};
+            WHERE {page_key} BETWEEN {pageset[0]} AND {pageset[1]}
+            {"ORDER BY " + page_key if sort_output else ""};
```
gwaybio, line 275: Is this ordering different than the `convert` ordering?

cytotable/utils.py
```diff
@@ -600,3 +570,54 @@ def evaluate_futures(sources: Union[Dict[str, List[Dict[str, Any]]], str]) -> An
...
+            Size of each chunk/page.
+
+    Returns:
+        List[Tuple[Union[int, float], Union[int, float]]]:
+            List of (start_key, end_key) tuples representing each page.
+    """
+
+    # if we have no keys, raise an exception
+    if not keys:
+        raise CytoTableException("No keys were provided to pageset generation process.")
```
gwaybio, line 594: Can you switch to active voice? Would a user provide the keys? Would some function generate them?

docs/source/overview.md
```diff
@@ -286,3 +286,16 @@ Also see CytoTable's presets found here: :data:`presets.config <cytotable.preset
 
 Note: data software outside of CytoTable sometimes makes use of the term "merge" to describe capabilities which are similar to join (for ex. [`pandas.DataFrame.merge`](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.merge.html).
 Within CytoTable, we opt to describe these operations with "join" to avoid confusion with software development alongside the technologies used (for example, [DuckDB SQL](https://duckdb.org/docs/archive/0.9.2/sql/introduction) includes no `MERGE` keyword).
+
+## Pagination
+
+CytoTable uses keyset pagination to help manage system-specific reasonable memory usage when working with large datasets.
+[Pagination](https://en.wikipedia.org/wiki/Pagination), sometimes also called paging or "data chunking", allows CytoTable to avoid loading entire datasets into memory at once while accomplishing tasks.
+Keyset pagination leverages existing column data as _pagesets_ to perform data extractions which focus on only a subset of the data as defined within the _pageset_ keys.
```
gwaybio, line 294: How are pageset keys defined?

gwaybio: Ah, I see the example below. It might be super simple to add thi:

"as defined within the _pageset_ keys (see example usage below)."

docs/source/overview.md
```diff
@@ -286,3 +286,16 @@ Also see CytoTable's presets found here: :data:`presets.config <cytotable.preset
...
+We use keyset pagination to reduce the overall memory footprint during extractions where other methods inadvertently may not scale for whole dataset work (such as offset-based pagination, which extracts then drops the offset data)([see here for more information](https://use-the-index-luke.com/no-offset)).
+
+```{eval-rst}
+Keyset pagination definitions may be defined using the ``page_keys`` parameter: :code:`convert(..., page_keys={"table_name": "column_name" }, ...)` (:mod:`convert() <cytotable.convert.convert>`).
+The ``page_keys`` parameter expects a dictionary where the keys are names of tables and values which are columns to be used for the keyset pagination pages.
+Pagination is created in conjunction with the ``chunk_size`` parameter which indicates the size of each page.
```
gwaybio, line 300: Can you add some comment on defaults? A user would only need to change this...when?

gwaybio (approved):
Exciting to see this enhancement! I made several comments, mostly aimed at improving clarity in comments and variable names. After addressing, feel free to merge 👍

cytotable/convert.py
d33bs: Thanks, good thought - I'll get this updated!

cytotable/convert.py
d33bs: Thanks - will update this as well!

cytotable/convert.py
d33bs: Thanks, great thought - this was a vestige of the earlier offser-based pagination and I believe we could update here for more performant operations as you mentioned.

cytotable/convert.py
d33bs: Thanks - will update as mentioned in the prior comment repsonse.

cytotable/convert.py
d33bs: Thanks - I'll add more context here and in the exception. In order for keyset pagination to operate the column / key name must be in the table. This exception raises if the column is not found in the table.

cytotable/convert.py
d33bs: Thanks - I'll add additional context as you suggested. Because the join composite table may include variable columns contingent on the join SQL there is a special key which may be specified in CytoTable to help perform keyset pagination on the join table. If there is no 'join' key specified when attempting to join we except here to help prevent further operations.

cytotable/utils.py
d33bs: Thanks, that's more clear - I'll update this accordingly.

cytotable/utils.py
d33bs: Thanks, yes, this is a "set" of the overall "whole" of all keysets for a given table. I'll add more context here to the documentation.

cytotable/utils.py
d33bs: Thanks for checking on this - this isn't different from the existing `convert` ordering but was restated to help with SQLite-specific processing here.

d33bs: I'm going to move this around a bit in coordination with other comments about sorting.

cytotable/utils.py
d33bs: Thanks, yes, I'll switch this to active voice. At the moment keys must be provided by a user via the `page_keys` parameter and eventually reach this function to help create "pagesets" which may be processed independently of one another.

docs/source/overview.md
d33bs: Thanks for this - I'll add further content as you mentioned to help clarify things.

docs/source/overview.md
d33bs: Thanks, I'll add further content here about when / why this might be changed.

cytotable/utils.py
d33bs: Looking into this more I realized it might not be possible to reach this exception case given the design of the function and how `convert` works. I ended up removing it instead to keep things simple.

cytotable/convert.py
d33bs: After looking into this more I felt that it'd be best to leave this line as-is and add additional changes to ensure the sorted results arrive sorted within the concatenation phase of CytoTable. I opted to do this because sorting for individual pages vs entirety of the dataset will have a lower performance impact and be more likely to succeed alongside resource constraints.

The result here will sort each pageset meaning the results within a single range of values are sorted. Afterwards, during concatenation to a single file, the results are combined using a natural sort over the filenames (which include the pageset range) to create a fully sorted dataset per source table and eventually joined table.

cytotable/convert.py
d33bs: Please reference https://github.com/cytomining/CytoTable/pull/220#discussion_r1731861318 for full description of how this was addressed.

cytotable/utils.py
d33bs: This was retained in coordination with this answer: https://github.com/cytomining/CytoTable/pull/220#discussion_r1731861318

d33bs:
Thank you @gwaybio, @jenna-tomkinson and @falquaddoomi for your reviews! After making adjustments for all comments I believe this is now ready and I'll merge this in.

PR #296, opened 2025-02-05
Update dependabot to group Python updates together

d33bs:
This PR updates dependabot to group updates together so they may be applied in one PR when possible. Dependencies which don't align are normally updated individually.


gwaybio (approved):

d33bs:
Thanks @gwaybio !

PR #310, opened 2025-04-08
Add DeepProfiler Compatibility

### LINKED ISSUE #15: Add DeepProfiler .npz as input for conversion to other formats
Add ability for pycytominer-transform to convert .npz to other formats (e.g. parquet, etc).

d33bs:
# Description

This PR adds DeepProfiler compatibility to CytoTable with necessary adjustments to existing functions.

Thanks for any feedback!

Closes #15 

# Checklist


cytotable/convert.py
```diff
@@ -399,17 +417,37 @@ def _source_pageset_to_parquet(
...
+    # See below for CSV and SQLite handling.
+    if source_type == ".npz":
+        return _extract_npz_to_parquet(
+            source_path=str(source["source_path"]),
+            dest_path=f"{source_dest_path}/{str(source['source_path'].stem)}.parquet",
+            tablenumber=source["tablenumber"],
+        )
+
+    if pageset is None:
+        # if we have a `None` pagest tand we're not using
```
gwaybio, line 444:
```suggestion
        # if we have a `None` pageset and we're not using
```

cytotable/convert.py
```diff
@@ -399,17 +417,37 @@ def _source_pageset_to_parquet(
...
+        # npz, then we have an exception (this shouldn't happen
+        # because we will need a pageset range to work with for
+        # table queries and npz files are handled above with
+        # the none case).
+        raise CytoTableException("No pageset range provided for source data.")
```
gwaybio, line 449: should this be clarified to state "for source .npz data"?

cytotable/convert.py
```diff
@@ -1001,94 +1039,86 @@ def _concat_join_sources(
...
     data_type_cast_map: Optional[Dict[str, str]] = None,
-) -> List[Tuple[str, str]]:
+) -> List[Tuple[str, pa.DataType]]:
     """
-    Infers a common schema for group of parquet files which may have
+    Infers a common schema for a group of parquet files which may have
     similar but slightly different schema or data. Intended to assist with
     data concatenation and other operations.
-
-    Args:
```
gwaybio, line 1010: wondering why this was deleted?

cytotable/presets.py
```diff
@@ -317,6 +317,33 @@
         # and modified at runtime as needed
         "CONFIG_JOINS": "",
     },
+    "deepprofiler": {
+        # version specifications using related references
+        "CONFIG_SOURCE_VERSION": {
+            "deepprofiler": "v0.3.1",
+            "cellprofiler": "v4.2.x",
```
gwaybio, line 324: why specified?

cytotable/presets.py
```diff
@@ -317,6 +317,33 @@
...
+        },
+        # names of source table compartments (for ex. cells.csv, etc.)
+        "CONFIG_NAMES_COMPARTMENTS": tuple(),
```
gwaybio, line 327: you might consider adding why these are not needed?

cytotable/presets.py
```diff
@@ -317,6 +317,33 @@
...
+        # names of source table metadata (for ex. image.csv, etc.)
+        "CONFIG_NAMES_METADATA": tuple(),
+        # column names in any compartment or metadata tables which contain
+        # unique names to avoid renaming
+        "CONFIG_IDENTIFYING_COLUMNS": tuple(),
+        # pagination keys for use with this data
+        # of the rough format "table" -> "column".
+        # note: page keys are expected to be numeric (int, float)
+        "CONFIG_PAGE_KEYS": {
+            "join": "Metadata_Site",
```
gwaybio, line 337: hmm, for single cells, isn't there an additional identifier?

cytotable/convert.py
d33bs: Thanks!

cytotable/convert.py
d33bs: Good thinking, thanks! I'll add that in.

cytotable/presets.py
d33bs: Based on my understanding from the input data to DeepProfiler CellProfiler may play a part when it comes to DeepProfiler's output. Since this field is largely information-based only (not used in processing currently) I felt it might be good to add the detail here based on where I thought the data might have come from (I used the dates of the commits for the files in Pycytominer as a basis of understanding what most likely was used).

cytotable/presets.py
d33bs: Thanks - will do!

cytotable/presets.py
d33bs: Where we could normally use object numbers I didn't find this within the data and so felt this might be the next best thing. This might be a gap in my understanding (I couldn't find a distinct way to relate the object numbers back to the table, mostly because they didn't exist within the NPZ from what I saw).

d33bs: In the case of the test data, I believe we're constricted to this index file's columns (which then go on to be included within the NPZ metadata): https://github.com/cytomining/pycytominer/blob/main/tests/test_data/DeepProfiler_example_data/inputs/metadata/test_index.csv .

d33bs:
Thanks @gwaybio ! I made some changes based on your comments. I left one comment open which I hope to double check on before we move forward with a merge here.

cytotable/presets.py
d33bs: Thanks @gwaybio ! After you mentioned this I realized I hadn't fully matched how location and feature data are treated when exported from NPZ to Parquet through this PR (these were single columns with array values). I've now shifted this to match how Pycytominer treats them, as new columns per array value (and changed the tests associated with this).

tests/test_convert_threaded.py
```diff
@@ -283,3 +284,89 @@ def test_avoid_na_row_output(
...
+        "Metadata_Model": "string",
+        "Location_Center_X": "double",
+        "Location_Center_Y": "double",
+    }
+    # check feature data types (we should only see double types)
+    assert {
+        str(field.type) for field in test_result.schema if "efficientnet_" in field.name
+    } == {"double"}
+
+    # check that we can use the resulting data with Pycytominer
```
gwaybio, line 353: 🎉

tests/test_convert_threaded.py
```diff
@@ -283,3 +284,89 @@ def test_avoid_na_row_output(
...
+                if "efficientnet_" in column
+            ],
+            image_features=False,
+            meta_features="infer",
+            method="standardize",
+            samples="all",
+            output_file="test_deepprofiler_normalized.parquet",
+            output_type="parquet",
+        )
+        is None
```
gwaybio, line 371: Consider adding a `.shape` assertion after loading in `test_deepprofiler_normalized.parquet` - you could even load it in with [`pycytominer.cyto_utils.load.load_profiles`](https://github.com/cytomining/pycytominer/blob/fd3d7664295be74c52e21bf31327619c99e10674/pycytominer/cyto_utils/load.py#L73) 🤷 (if that works, not sure if it even does anymore)

gwaybio (approved):

cytotable/convert.py
```diff
@@ -360,14 +376,16 @@ def _get_table_keyset_pagination_sets(
...
+    # wherein it considers _generate_pagesets to be invalid.
+    return _generate_pagesets(page_keys, chunk_size)  # type: ignore[return-value]
 
 
 @python_app
 def _source_pageset_to_parquet(
     source_group_name: str,
     source: Dict[str, Any],
-    pageset: Tuple[Union[int, float], Union[int, float]],
+    pageset: Optional[Tuple[Union[int, float], Union[int, float]]],
```
falquaddoomi, line 388: I'm not sure this type covers `_get_table_keyset_pagination_sets()` now returning `List[None]`; I believe `Optional[...]` covers just it returning `None`.

cytotable/sources.py
```diff
@@ -93,16 +95,22 @@ def _get_source_filepaths(
...
+                targets is not None
+                and str(subpath.stem).lower() in [target.lower() for target in targets]
+                or subpath.suffix.lower() == ".sqlite"
+            )
+            # Otherwise, include files matching the source_datatypes
+            or (
+                targets is None
+                or targets == []
+                and subpath.suffix.lower() in source_datatypes
+            )
```
falquaddoomi, line 109: This is a lot clearer than before; kudos!

cytotable/convert.py
```diff
@@ -399,17 +417,42 @@ def _source_pageset_to_parquet(
...
+    # If we have npz files, we need to extract them in a specialized manner.
+    # See below for CSV and SQLite handling.
+    if source_type == ".npz":
+        return _extract_npz_to_parquet(
+            source_path=str(source["source_path"]),
+            dest_path=f"{source_dest_path}/{str(source['source_path'].stem)}.parquet",
+            tablenumber=source["tablenumber"],
+        )
+
+    if pageset is None:
```
falquaddoomi, line 443: Kind of a silly thing, but maybe make this an `elif` chained to the above `if`, so it's obvious that this is related to it not being an NPZ? It's functionally the same, but might be more readable if someone's skimming and didn't see the return above.

falquaddoomi (commented):
Nice work! This is a pretty complex addition, and it looks like you approached it in a structured way, so kudos for that. I left a few comments mostly about typing and style.

docs/source/overview.md
```diff
@@ -78,6 +78,12 @@ A quick example of how this argument is used: :code:`convert(..., local_cache_di
 
 Future work to enable direct SQLite data access from cloud locations for CytoTable will be documented within GitHub issue [CytoTable/#70](https://github.com/cytomining/CytoTable/issues/70).
 
+__NumPy Zipped Archive (.npz)__:
+
+CytoTable downloads cloud-based NPZ archives locally before other CytoTable processing.
```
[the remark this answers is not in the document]
d33bs: Good idea! I'll add that in, thanks!

cytotable/convert.py
```diff
@@ -279,7 +286,9 @@ def _get_table_keyset_pagination_sets(
     page_key: str,
     source: Optional[Dict[str, Any]] = None,
     sql_stmt: Optional[str] = None,
-) -> Union[List[Tuple[Union[int, float], Union[int, float]]], None]:
+) -> Union[
+    List[Optional[Tuple[Union[int, float], Union[int, float]]]], List[None], None
```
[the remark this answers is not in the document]
d33bs: Thanks good points! I'll see about addressing these.

cytotable/convert.py
```diff
@@ -399,17 +417,42 @@ def _source_pageset_to_parquet(
...
+        # if we have a `None` pageset and we're not using
+        # npz, then we have an exception (this shouldn't happen
+        # because we will need a pageset range to work with for
+        # table queries and npz files are handled above with
+        # the none case).
+        raise CytoTableException(
+            (
+                "No pageset range provided for source data"
+                " (applies to non-NPZ datasets)."
+            )
```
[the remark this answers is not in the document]
d33bs: Dig it - I'll add this in.

tests/test_convert_threaded.py
d33bs: Added in a further commit.

cytotable/sources.py
d33bs: Thanks, agreed!

cytotable/convert.py
d33bs: Makes sense to me, thanks! I agree this is clearer - I've gotten into a habit lately of using `elif` less and less but this is a spot where it would indeed increase readability.

cytotable/convert.py
d33bs: Thanks for raising this - I think the key here is that this is a "set" and not "sets"; e.g. `[None]` or `[(0,0)]` for "sets" and `None` or `(0,0)` for a "set". I'm going to leave this as-is as a result.

d33bs:
Thank you @gwaybio and @falquaddoomi for your reviews! After incorporating changes based on your comments into the code I'll now merge this in.

PR #441, opened 2026-04-28
Follow symlinked directories in local source discovery (fixes #440)

### LINKED ISSUE #440: convert(): symlinked input directories silently yield NoInputDataException (pathlib.Path.glob doesn't follow symlinks)
## Summary

`cytotable.convert(source_path=<dir>, source_datatype=\"csv\", ...)` silently fails with `NoInputDataException` when CSVs live inside a directory that is reachable only through a symlinked subdirectory, even though every file is plainly visible to `find`/`os.walk(followlinks=True)`. Root cause: `pathlib.Path.glob(\"**/*\")` does not follow symlinks (default `recurse_symlinks=False` on 3.13+; never followed on 3.10–3.12).

This breaks any Nextflow workflow that hands CytoTable a directory of staged inputs, because Nextflow's default `stageInMode` symlinks input directories into the task work dir.

## Affected versions

- Reproduced on **CytoTable 0.0.15**, **Python 3.13.3** (Linux, container `community.wave.seqera.io/library/pip_cytotable:e5e76f6f7c7bea96`).
- The same pattern is present on `main` (commit `a957113b9583`).

## Where the bug lives

**v0.0.15** — single offending site:
- `cytotable/sources.py:94` — `(x for x in path.glob(\"**/*\") if x.is_file())`

**main branch** — refactored into `cloud_glob()` in `cytotable/utils.py`, but the same `Path.glob` pattern repeats in three local-filesystem branches:
- `cytotable/utils.py:1011` — non-S3 CloudPath fallback
- `cytotable/utils.py:1021` — local `pathlib.Path` branch
- `cytotable/utils.py:1045` — string path → `Path` branch

All three call `Path.glob(pattern)` without `recurse_symlinks=True`.

## Minimal reproduction

```bash
# 1. Tree with a symlinked analysis subdir (mimics Nextflow stageInMode=symlink)
mkdir -p /tmp/cytotable_symlink_bug/real/analysis
mkdir -p /tmp/cytotable_symlink_bug/staged/1
printf 'ImageNumber,Metadata_Plate,Metadata_Well,Metadata_Site\n1,Plate1,A01,1\n' \
  > /tmp/cytotable_symlink_bug/real/analysis/Image.csv
for f in Cells Cytoplasm Nuclei; do
  printf 'ImageNumber,ObjectNumber\n1,1\n' > /tmp/cytotable_symlink_bug/real/analysis/$f.csv
done
ln -s ../../real/analysis /tmp/cytotable_symlink_bug/staged/1/analysis

# 2. Run repro
docker run --rm -v /tmp/cytotable_symlink_bug:/data \
  community.wave.seqera.io/library/pip_cytotable:e5e76f6f7c7bea96 \
  python - <<'PY'
import pathlib, sys
from importlib.metadata import version
from cytotable.sources import _get_source_filepaths
from cytotable.exceptions import NoInputDataException

print(\"Python:\", sys.version.split()[0], \"| CytoTable:\", version(\"cytotable\"))

for label, p in [(\"real (no symlink)\", \"/data/real\"),
                 (\"staged (symlinked subdir)\", \"/data/staged\")]:
    try:
        r = _get_source_filepaths(path=pathlib.Path(p), source_datatype=\"csv\")
        print(f\"{label}: OK, {sum(len(v) for v in r.values())} files\")
    except NoInputDataException as e:
        print(f\"{label}: NoInputDataException: {e}\")
PY
```

### Observed output

```
Python: 3.13.3 | CytoTable: 0.0.15
real (no symlink): OK, 4 files
staged (symlinked subdir): NoInputDataException: No input data to process at path: /data/staged
```

### Expected output

Both invocations should discover the four CSVs, since they are reachable in both layouts (`find -L /data/staged -name '*.csv'` lists all four).

## Why this matters

Pipelines that orchestrate per-site CellProfiler runs and then call `cytotable.convert` on a parent directory of staged outputs are the canonical use case for the `cellprofiler_csv` preset. Nextflow stages each per-site directory as a symlink by default — so today the only workarounds are:

1. Set `stageInMode 'copy'` on the calling process (doubles I/O for plates with hundreds of sites), or
2. Resolve symlinks in user code before invoking CytoTable.

Both shift the burden to every downstream consumer of CytoTable and silently regress when users don't realize symlinks are involved.

### v0.0.15 (`cytotable/sources.py:94`) — Python 3.13 only

```python
else (x for x in path.glob(\"**/*\", recurse_symlinks=True) if x.is_file())
```

### main branch — backward-compatible helper (CytoTable supports 3.10–3.14)

`recurse_symlinks` is only available on Python 3.13+. A small helper covers the older versions:

```python
import os, sys
from pathlib import Path

def _glob_follow_symlinks(path: Path, pattern: str):
    \"\"\"Recursive glob that follows symlinked directories on all supported Pythons.\"\"\"
    if sys.version_info >= (3, 13):
        yield from path.glob(pattern, recurse_symlinks=True)
    else:
        for root, dirs, files in os.walk(str(path), followlinks=True):
            root_path = Path(root)
            for name in dirs + files:
                yield root_path / name
```

Then replace each `path.glob(\"**/*\")` / `start.glob(pattern)` / `base.glob(pattern)` in the local-filesystem branches of `cytotable/utils.py` with `_glob_follow_symlinks(...)`.

A regression test that builds the symlink layout above and asserts a non-empty Parquet would lock this in.

## Related context

- Searched the issue tracker for \"symlink\", \"follow_symlinks\", \"glob\", \"Nextflow\" — no existing issue covers this.
- Encountered while implementing per-plate CytoTable invocation in [nf-core/cellpainting](https://github.com/nf-core/cellpainting) (issue [#40](https://github.com/nf-core/cellpainting/issues/40)).

kenibrewer:
## Summary

`cytotable.convert()` was silently raising `NoInputDataException` whenever its input directory contained data reachable only through a symlinked subdirectory — the default layout produced by Nextflow's `stageInMode='symlink'`. Root cause: `pathlib.Path.glob` does not follow directory symlinks before Python 3.13.

This PR adds a small `_glob_follow_symlinks(start, pattern)` helper and uses it in the local-path branches of `cloud_glob`. On Python 3.13+ it delegates to `Path.glob(pattern, recurse_symlinks=True)`. On 3.10–3.12 it walks the tree with `os.walk(..., followlinks=True)` and applies the post-`**/` portion of the pattern at each visited directory, which covers the patterns `cloud_glob` actually uses (`**/*`, `**/*.csv`, etc.).

The non-S3 `CloudPath` branch is left alone — real cloud providers don't have filesystem-level symlinks, and `cloudpathlib.CloudPath.glob` doesn't accept `recurse_symlinks` anyway.

Fixes #440.

## Test plan

- [x] New regression test `tests/test_utils.py::test_cloud_glob_follows_symlinked_directories` builds the staged-symlink layout from the issue and asserts `cloud_glob` returns the four CSVs for both `Path` and `str` inputs. Confirmed it fails before the helper change and passes after.
- [x] All existing `tests/test_utils.py` and `tests/test_sources.py` tests still pass (17 passed locally on Python 3.12.7).
- [x] Re-ran the issue's exact reproduction against `_get_source_filepaths`: `staged (symlinked subdir)` now returns 4 files (was raising `NoInputDataException`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved filesystem globbing: now follows symlinked directories, deduplicates identical targets reachable via multiple symlinks, and prevents infinite traversal on cyclic symlinks; preserves expected glob semantics for anchored and mid-pattern usage.

* **Tests**
  * Added regression tests covering symlink traversal, anchored and mid-pattern glob semantics, deduplication, and cycle termination.

* **Chores**
  * Bumped the mypy pre-commit hook to a newer version.

### DIFF
--- a/cytotable/utils.py
+++ b/cytotable/utils.py
@@ -5,6 +5,7 @@ Utility functions for CytoTable
 import logging
 import os
 import pathlib
+import sys
 from pathlib import Path, PurePosixPath
 from typing import Any, Dict, Iterator, List, Optional, Tuple, Union, cast
 
@@ -923,6 +924,26 @@ def find_anndata_metadata_field_names(
     return numeric_fields, non_numeric_fields
 
 
+def _glob_follow_symlinks(start: Path, pattern: str) -> Iterator[Path]:
+    """
+    Like ``Path.glob(pattern)``, but follows symlinked directories on every
+    Python version CytoTable supports. ``Path.glob`` only gained
+    ``recurse_symlinks=True`` in 3.13; on 3.10–3.12 we walk the tree with
+    ``os.walk(followlinks=True)`` and apply the post-``**/`` portion of the
+    pattern at each visited directory.
+    """
+    if sys.version_info >= (3, 13):
+        yield from start.glob(pattern, recurse_symlinks=True)
+        return
+
+    if pattern.startswith("**/"):
+        local_pattern = pattern[3:]
+        for root, _dirs, _files in os.walk(str(start), followlinks=True):
+            yield from Path(root).glob(local_pattern)
+    else:
+        yield from start.glob(pattern)
+
+
 def cloud_glob(
     start: Union[str, CloudPath, Path],
     pattern: str,
@@ -1018,7 +1039,7 @@ def cloud_glob(
     # ---- Branch 2: local Path ----
     if isinstance(start, Path):
         yielded = 0
-        for child in start.glob(pattern):
+        for child in _glob_follow_symlinks(start, pattern):
             yield child
             yielded += 1
             if max_matches is not None and yielded >= max_matches:
@@ -1042,7 +1063,7 @@ def cloud_glob(
         else:
             base = Path(start)
             yielded = 0
-            for child in base.glob(pattern):
+            for child in _glob_follow_symlinks(base, pattern):
                 yield child
                 yielded += 1
                 if max_matches is not None and yielded >= max_matches:
--- a/cytotable/utils.py
+++ b/cytotable/utils.py
@@ -931,6 +931,18 @@ def _glob_pattern_matches(rel_parts: Tuple[str, ...], pat_parts: List[str]) -> b
     semantics: ``**`` matches zero or more components, ``*`` and ``?`` are
     fnmatch wildcards within a single component, and matching is anchored at
     the left of ``rel_parts``.
+
+    Args:
+        rel_parts:
+            Path components of the candidate, relative to the search root
+            (e.g. ``("analysis", "Cells.csv")``).
+        pat_parts:
+            Pattern components produced by splitting the glob on ``"/"``
+            (e.g. ``["**", "*.csv"]``).
+
+    Returns:
+        ``True`` if ``rel_parts`` matches ``pat_parts`` under the semantics
+        described above, else ``False``.
     """
     if not pat_parts:
         return not rel_parts
@@ -957,6 +969,17 @@ def _glob_follow_symlinks(start: Path, pattern: str) -> Iterator[Path]:
     tree with ``os.walk(followlinks=True)`` and match each entry's relative
     path against the full pattern, so any pattern accepted by ``Path.glob``
     works across versions.
+
+    Args:
+        start:
+            Root directory to glob under. Must reference a local or network
+            filesystem path.
+        pattern:
+            Glob pattern relative to ``start`` (e.g. ``"**/*.csv"``).
+
+    Yields:
+        ``pathlib.Path`` entries matching ``pattern``, deduplicated so that
+        two paths resolving to the same real file are only yielded once.
     """
     if sys.version_info >= (3, 13):
         # ``Path.glob(recurse_symlinks=True)`` yields each reachable path,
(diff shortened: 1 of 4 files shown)


cytotable/utils.py
```diff
@@ -923,6 +925,79 @@ def find_anndata_metadata_field_names(
     return numeric_fields, non_numeric_fields
 
 
+def _glob_pattern_matches(rel_parts: Tuple[str, ...], pat_parts: List[str]) -> bool:
+    """
+    Match path components against pattern components using pathlib-glob
+    semantics: ``**`` matches zero or more components, ``*`` and ``?`` are
+    fnmatch wildcards within a single component, and matching is anchored at
+    the left of ``rel_parts``.
+    """
```
d33bs, line 946: Consider completing this docstring with parameters and return types, in alignment with the rest of the project. Similar comments for other new functions added (including if there is no docstring added).

tests/test_utils.py
d33bs, line 1: Consider compacting these tests using pytest parameterize to avoid duplication of code where possible and reasonable (without losing ground on the coverage).

d33bs (changes requested):
Nice! I think this is pretty close to a merge. In general I think we should also be upfront about our handling of paths somehow within the sphinx docs, perhaps under https://cytomining.github.io/CytoTable/architecture.technical.html#data-paths (with the changes here I think a sentence or two could help elaborate a bit).

cytotable/utils.py
kenibrewer: Addressed in ef51015 — added Args/Yields/Returns sections to `_glob_pattern_matches`, `_glob_follow_symlinks`, and `_walk_and_match` to match the project's existing docstring style.

tests/test_utils.py
kenibrewer: Addressed in 90a0394 — collapsed the symlink-discovery and non-`**/`-prefix tests into parametrized forms. The dedup and cycle tests are kept separate because they exercise distinct layouts and concepts.

cytotable/utils.py
kenibrewer: Thoughts on using pydocstyle to enforce this convention?

kenibrewer: Opened a PR to enforce this: https://github.com/cytomining/CytoTable/pull/443

kenibrewer:
Addressed requested changes, and opened a separate [PR](https://github.com/cytomining/CytoTable/pull/443) to enforce pydocstyle conventions.

d33bs (approved):
Thanks @kenibrewer - this looks great!

